PR #1, opened 2020-02-02
Resposive-cake-webpage

mahmutsel:


LucyMac:
Hi Mahmut,
I hope you've managed to link your CSS file to your project 🙂 
Can you push your latest changes?
Also, please change the name of your PR to describe what you're making.
Thanks

mahmutsel:
Hi Lucy, I have just completed some part of homework. There is a lot more work to do but before class at least i can push this one. thanksss :)

style.css
```diff
@@ -1 +1,157 @@
...
+    .header-brand{ 
+        margin: 0;
+        text-align:left;
+        padding:0px 20px 0px 30px;
+        border-right:3px solid #111;
+        float:left;
+    }
+    header nav ul {
+        margin:20px 0px 0px 20px;
+        float:right;
```
LucyMac: Don't use floats - use flexbox instead!

style.css
```diff
@@ -1 +1,157 @@
...
+
+}
+header nav ul {
+    display:block;
+    margin:0 auto;
+    width:fit-content;
+}
+header nav ul li {
+    display:inline-block;
+    float:left;
```
LucyMac: Don't use floats - use flexbox instead!

style.css
```diff
@@ -1 +1,157 @@
...
+        height: 400px;
+        background-image:url(https://imagesvc.meredithcorp.io/v3/mm/image?url=https%3A%2F%2Fimg1.cookinglight.timeinc.net%2Fsites%2Fdefault%2Ffiles%2Fstyles%2Fmedium_2x%2Fpublic%2F1542062283%2Fchocolate-and-cream-layer-cake-1812-cover.jpg%3Fitok%3DrEWL7AIN);
+        background-repeat:no-repeat;
+        background-position:center;
+        background-size:cover
+        display:center;
+    }
+    .index-banner h1{
+        display:center;
+        width:1000px;
```
LucyMac: This is breaking your page layout because the rule applies even when the screen size is smaller than 1000px. 
Can you remove it and make the layout work using flexbox?

mahmutsel:
Hi Lucy, I removed the 1000px scren size, it is better now. thanks. But I am not very succesfull at converting float to flexbox. I think i am missing some points but  I couldnt find it where the problem is.

PR #3, opened 2020-02-02
Cake project - Rayhan

Rayhan94:


index.html
```diff
@@ -1,16 +1,105 @@
...
+            <!-- Add a link to your css file here -->
+        </head>
 
-<body>
-    <!-- Add your markup here -->
-</body>
+        <body>
+        <main>
+        <!--navigation and logo-->
+        <nav class="upper-navigation">
```
amcraethanan: remember to make use of the `<header>` element, the nav element here should be nested within there

index.html
```diff
@@ -1,16 +1,105 @@
...
+        </head>
 
-<body>
-    <!-- Add your markup here -->
-</body>
+        <body>
+        <main>
+        <!--navigation and logo-->
+        <nav class="upper-navigation"> 
+            <header><center>Welcome to Sweet cakes!</center></header>
```
amcraethanan: `<center>` is not a valid html element - have a quick read on how to use the [header element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/header)

index.html
```diff
@@ -1,16 +1,105 @@
...
+                <li>Home</li>
+                <li>Cakes</li>
+                <li>Ordering</li>
+                <li>Lessons</li>
+                <li>About</li>
+            </ul>
+        
+        </nav>
+        <img class="cakeLogo" src="images/cake-logo.jpg" alt="logo" width="150px" height="150px">
+
```
amcraethanan: remember to make use of the `<section>` html to seperate out your page into logical sections

index.html
```diff
@@ -1,16 +1,105 @@
...
 
-</html>
\ No newline at end of file
+            <section class="carousel">
+                    <!--making the image width 100% to make it responsive-->
+            <ul>
+                    
+                <li>
+                    <div class="carousel-container">
+                        <img src="images/cake.jpg" alt="Snow" style="width:100%">
```
amcraethanan: Remeber to not use inline styles within a html element - instead make use of css classes

index.html
```diff
@@ -1,16 +1,105 @@
...
+                    </li>
+                  </ul>
+                </div>
+              </section>
+
+            <!--Footer for sitemap etc...-->
+            <footer>
+                <nav>
+                    <ul class="footer-nav">
+                    <li><a href="#">Sitemap</a></li>
```
amcraethanan: Remember to indent elements that are children of other elements to  make it easier to locate start and end tags when coding

index.html
```diff
@@ -1,16 +1,105 @@
...
+                    <li><a href="#">Contact</a></li>
+                    <li><a href="#">Delivery</a></li>
+                    </ul>
+                </nav>
+            </footer>
+        </main>
```
amcraethanan: The footer html element should appear after the `<main>` element - have a quick read to remind yourself of the [main](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main) and [footer](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/footer) html elements

style.css
```diff
@@ -1 +1,179 @@
 /*  Add your styling here */
+@media(max-width: 1024px){
```
amcraethanan: media queries should live at the bottom of the stylesheet - CSS is read from top to bottom; this means that any styles defined in this block will now get overwritten by any subsequent styles

style.css
```diff
@@ -1 +1,179 @@
...
+    
+  
+}
+
+body{
+    background-color: pink;
+}
+
+.upper-navigation {
+    justify-content: space-around;
```
amcraethanan: in order for this to have any affect, you will need to make use of `display: flex` - have a look at last weeks lesson to look over flexbox and how this is used to remind yourself

style.css
```diff
@@ -1 +1,179 @@
...
+    background-color: rgb(217, 236, 252);
+    position: relative;
+    margin: 0 auto;
+    
+  }
+  
+  .upper-navigation {
+    width: 970px;
+    height: 4.375rem;
```
amcraethanan: instead of using explicit heights and widths, think about implementing flexbox and use percentages so that the sizing is relative to the screen widths

style.css
```diff
@@ -1 +1,179 @@
...
+    margin: 0 auto;
+    color: #f721da;
+    justify-content: center;
+   
+  }
+  
+  .upper-navigation{
```
amcraethanan: this is now the third instance of this css class - choose one definition and keep it, remember that css is read from top to bottom, so only the last definition of the css class will be applied

style.css
```diff
@@ -1 +1,179 @@
...
+  
+  .upper-navigation li {
+    display: inline-block;
+    margin-left: 75px;
+  }
+
+  .cakeLogo{
+    position:absolute;
+    left:0;
+    top:0;
```
amcraethanan: what about applying padding or margin to this so that it is not flush against the edge of the browser?

style.css
```diff
@@ -1 +1,179 @@
...
+  }
+
+
+.intro{
+    display: inline-block;
+    margin-left:
```
amcraethanan: why is this missing a value?

style.css
```diff
@@ -1 +1,179 @@
...
+    justify-content: space-between;
```
amcraethanan: this is only applied when `display` is set to `flex`

style.css
```diff
@@ -1 +1,179 @@
...
+    overflow: hidden;
+    margin-top: 100px;
+    box-shadow: 10px 10px 15px rgba(0,0,0,0.50);
+    
+}
+
+.carousel li{
+    display: inline-block;
+    height: 300px;
+    float: left;
```
amcraethanan: instead of using float, try to use flexbox

PR #5, opened 2020-02-02
hamid-cake

Hamidreza-Fakharzadeh:


Hamidreza-Fakharzadeh:
i appreciate it if you could check my flex box css and give some suggestion.

index.html
```diff
@@ -5,12 +5,81 @@
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <link rel="stylesheet" href="//fonts.googleapis.com/css?family=Open+Sans:300,400,600,700&amp;lang=en" />
```
AlfioEmanueleFresta: Here you are importing "Open Sans" from Google Fonts, but aren't using the font anywhere in the page. Did you mean to import "Dosis", which is used in your stylesheet?

style.css
```diff
@@ -1 +1,110 @@
...
+}
+*{ 
+    box-sizing:border-box;
+    font-family:"Dosis", sans-serif; 
+  }
+header{
+   display: flex;
+   flex-direction: column;
+   }
+  .menu{
```
AlfioEmanueleFresta: Make sure to use correct indentation in your HTML and CSS. Whitespace makes all the difference when reading someone else's code!

style.css
```diff
@@ -1 +1,110 @@
...
+   }
+   h2 {
+     text-align: center;
+   }
+   .picture-text {
+     display: flex;
+     flex-direction: column;
+     align-items: center;
+   }
+.pictureWithText1 {
```
AlfioEmanueleFresta: When referring to a single element, use IDs rather than classes, e.g.:

```html
<div class="picture-with-text" id="picture-1">...</div>
<div class="picture-with-text" id="picture-2">...</div>
...
```

```css
#picture-1 {
  background: url(...);
}

#picture-2 {
  background: url(...);
}
```

index.html
```diff
@@ -5,12 +5,81 @@
...
-    <!-- Add a link to your css file here -->
+    <link rel="stylesheet" type="text/css" href="style.css" />
+
 </head>
 
 <body>
-    <!-- Add your markup here -->
+    <header>
+        <nav class="main-navigation">
+          <ul class="menu">
```
AlfioEmanueleFresta: This looks complicated - I would recommend starting with a single navigation menu (rather than a `menu` and a `sub-menu`).

If you want to make it responsive, you can start with something simple, such as having the elements align vertically or horizontally.

style.css
```diff
@@ -1 +1,110 @@
...
+    }
+  .sub-menu {
+    display: flex;
+  }
+  .include-both-section {
+    display: flex;
+    flex-direction: row-reverse;
+  }
+  .slider {
+    display: flex;
```
AlfioEmanueleFresta: Within a media query, you should only include what you want to be *different* when using a smaller screen. You already have `display: flex` for the `.slider` outside of the media query, so you don't need to have this again here.

index.html
```diff
@@ -5,12 +5,81 @@
...
+          <h3>English Cake</h3>
+          <h5>relax time with nice English cake</h5>
+        </div>
+        <div class="picture-text pictureWithText3">
+          <h3>American cake</h3>
+          <h5>painting dream with nice American cake</h5>
+        </div>
+        <div class="picture-text pictureWithText4">
+          <h3>Rusha cake</h3>
+          <h5>play chees with nice Rusha</h5>
```
AlfioEmanueleFresta: `<h*>` tags are for titles / headings. If this is a description, consider using a different tag such as `<p>` (paragraph of text).

AlfioEmanueleFresta:
@Hamidreza-Fakharzadeh this is a good start! I've added a few comments above.

Hamidreza-Fakharzadeh:
Thank you for your checking my code. Now l am straggling with using carousel in flexbox . I would like to build up the page in a simple way as you suggested.

PR #6, opened 2020-02-03
website is done

AlexeyBrambalov:


style.css
```diff
@@ -1 +1,146 @@
...
+    width: 15wv;
+}
+
+.contacts{
+    background-color: darkgrey;
+    width: 20wv;
+}
+
+.nav__list{
+    display: flex;
```
amcraethanan: well done on remembering to use flexbox from last week's lesson 👍

index.html
```diff
@@ -7,10 +7,55 @@
...
+        </ul>
+    </nav>
+       
+    
+
+</header>
+
 <body>
-    <!-- Add your markup here -->
+    <div class="main__inner">
```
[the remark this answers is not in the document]
jamesgorrie: There's a [nice explanation here as to why this is important](https://www.w3.org/WAI/tutorials/page-structure/), not only for machines, but for humans too.

[The practical bits are explained here](https://www.w3.org/WAI/tutorials/page-structure/regions/).

style.css
```diff
@@ -1 +1,146 @@
-/*  Add your styling here */
+.header{
+    width: 100%;
+    top: 0;
+    left: 0;
+    right: 0;
```
jamesgorrie: These `0` values are not necessary as they are the default, they don't do any harm, but will make me wonder why you've used them if I look at them in the future...

style.css
```diff
@@ -1 +1,146 @@
...
+}
+
+.logo{
+    background-color: grey;
+    width: 15wv;
+}
+
+.contacts{
+    background-color: darkgrey;
+    width: 20wv;
```
jamesgorrie: Do you mean `vw`? In which case this will be massive.

style.css
```diff
@@ -1 +1,146 @@
...
+    display: flex;
+    flex-direction: row;
+    list-style: none;
+    justify-content: space-between;
+    align-items: stretch;
+    display: none;
+}
+
+.nav__item {
+    background-color: deeppink;
```
jamesgorrie: Great colour 🎨 !

style.css
```diff
@@ -1 +1,146 @@
...
+    }
+    .main__info{
+        width: 25vw;
+        height: 25vh;
+    }
+    .main__image{
+        width: 75vw;
+    }
+
+}
```
jamesgorrie: I have found that keeping my styles together makes it easier to work out what's going on... e.g.

```CSS
.nav__list {
  /* Desktop */
}
@media (min-width: 768px) {
  .nav__list {
    /* mobile css */
  }
}

.main__info {
  /* Desktop */
}
@media (min-width: 768px) {
  .main__info {
    /* mobile css */
  }
}
```

This makes it easier to see what is different between the two versions as I generally work on 1 section of a site at once.

You would also be able to spot that you have written a lot of the same CSS for both versions!

PR #7, opened 2020-02-05
Mobile website

banirezaie:
change on 04/feb/20
just navbar


LucyMac:
Awesome navigation! 💪 

Please continue to work on this and push your work so we can see your latest changes.

style.css
```diff
@@ -1 +1,194 @@
...
+    margin-right: 1em;
+}
+figcaption{
+    text-align: center;
+    text-shadow: saddlebrown;
+}
+p{
+    font-size: medium;
+}
+
```
ChrisOwen101: Remove the blank space here

style.css
```diff
@@ -1 +1,194 @@
...
+    height: 100%;
+    width: 100%;
+    position: absolute;
+    transition: all .2s ease-out;
+    top: -5px;
+}
+.header1 .menu-btn{
+    display: none;
+}
+.header1 .menu-btn:checked ~ .menu{
```
ChrisOwen101: You probably don't need the `1` after all of these names and it would easier to read if you didn't :)

Maybe try renaming them

style.css
```diff
@@ -1 +1,194 @@
...
+
+
+
+
+
+
+
+@media (min-width:768px){
+    .header1 li{
+        float: left;
```
LucyMac: Do't use floats please - use flexbox instead

style.css
```diff
@@ -1 +1,194 @@
...
+
+.header1{
+    grid-area: header;
+    background-color: rgb(235, 195, 123);
+    position: fixed;
+    box-shadow: 1px 1px 4px rgba(0,0,0,.2);
+    width: 100%;
+    z-index: 10;
+    height:75px;
+}
```
LucyMac: if you set this to use `display: flex`, you can set it to `flex-direction: column` on mobile, and then line the contents in a row on desktop with `flex-direction :row`

style.css
```diff
@@ -1 +1,194 @@
...
+.header1 ul a{
+    display: block;
+    /* outline: 1px solid purple; */
+    padding:15px;
+    border:1px solid purple;
+    text-decoration: none;
+}
+
+.header1 .logo{
+    float: left;
```
LucyMac: Remove floats from all your CSS, you shouldn't need it. Modern CSS makes use of Flexbox to replace this.

style.css
```diff
@@ -1 +1,194 @@
...
+    display: block;
+    margin-top: 5px;
+    margin-left: 5px;
+}
+.header1 ul a:hover{
+    background-color:rgb(226, 165, 52);
+    text-decoration: none;
+}
+.header1 .menu{
+    clear: both;
```
LucyMac: If you remove floats, you won't need this rule anymore.

style.css
banirezaie: ok, I did it.

style.css
banirezaie: ok, I did it.

style.css
banirezaie: I removed it.

style.css
banirezaie: I just removed it, it didn't need flex box as well.

style.css
banirezaie: but, it didn't  work with flex box...

style.css
banirezaie: it didn't work with : display: flex; flex-direction:row;
it still stay in column, I changed it anyway and make float a comment.

PR #8, opened 2020-02-05
Serkan cake webpage

SCavus:


ChrisOwen101:
This is looking pretty good on a phone screen size:
image
But not so good on a desktop:
image

It's important that it works in both!

PR #9, opened 2020-02-06
RC basic mobile version

279Roxana:


style.css
```diff
@@ -1 +1,365 @@
...
+
+.cakes {
+    display: flex;
+    flex-direction: column;
+    border: 3px solid black;
+    margin: 3%;
+    background-color: rgb(204, 154, 240);
+    height: 30%;
+    width: 99%;
+    overflow: hidden;
```
amcraethanan: consider adding padding so that the edge isn't as flush against the edge of the browser

style.css
```diff
@@ -32,15 +33,16 @@ header {
 
 .menu li {
     list-style-type: none;
-    padding: 0.5%;
+    padding: 11.5%;
```
LucyMac: In general I would avoid using % for padding and margin. It means you can't control exactly how big the spacing will be, which will give inconsistent gaps between elements. Usually, designers will have fixed pixel sizes that they use for all spacing to keep it consistent.

style.css
```diff
@@ -178,8 +181,8 @@ a:hover {
 /* images */
 
 .section2-picture {
-    height: 15%;
-    width: 70%;
+    height: auto;
+    width: 100%;
```
LucyMac: Good - always set _either_ the width _or_ the height of an image. Leaving the other to `auto` works perfectly to preserve the aspect ratio 👍

style.css
```diff
@@ -1 +1,384 @@
...
+    overflow: auto;
+    -webkit-overflow-scrolling: touch;
+    margin: 5%;
+    background-color: rgb(241, 224, 227);
+    border: 2px solid grey;
+    border-radius: 2%;
+    list-style-type: none;
+    height: 30%;
+    width: 70%;
+}
```
LucyMac: This `carousel` is a `ul` and all `ul`s are given a default `padding-left` by the browser. Add `padding-left: 0` here to remove this undesired spacing (and do this for all your `ul` elements in the future)

style.css
```diff
@@ -1 +1,384 @@
...
+
+    .carousel {
+       width: 80%;
+       height: auto;
+       border: 2px solid black;
+    }
+
+    .carousel-image {
+        height: 80%;
+        width: 90%;
```
LucyMac: You're setting a width AND a height here. This will distort your image. Set the width only, and leave the height to `auto`

PR #10, opened 2020-02-06
last one on thursday

abcelen:
Hi
I tried to make it look nice and neat as well.


index.html
```diff
@@ -7,10 +7,92 @@
...
+</header>
+<nav>  
+  <div class="topnav">
+    <a href="#">Home</a>
+    <a href="#">Cakes</a>
+    <a href="#">Ordering</a>
+    <a href="#">Lessons</a>
+    <a href="#">About</a>
+  </div>
+</nav>
```
amcraethanan: As the navigation typically is always at the top of the page, this nav bar could live in the header element.

amcraethanan: As the navigation typically is always at the top of the page, this nav bar could live in the header element.

index.html
```diff
@@ -7,10 +7,92 @@
...
+    <a href="#">About</a>
+  </div>
+</nav> 
+
+
+    <div class="row">
+        <div class="rightcolumn">
+            <div class="card">
+                <h2>CAROUSSEL</h2>
+                <div class="img" style="height:200px;">Image</div>
```
amcraethanan: Remember to put styling in the class rather than inline in the html element; this keeps all styling in one place. Should this also have a image element?

index.html
```diff
@@ -7,10 +7,92 @@
...
+                <p>Prefer to do it the old fashioned way?<br>
+                    We have stores all across London so why not pop in, say, 'Hi' and pick up a box of your favourites!
+                    Click here to find my nearest store/collection point <br></p>
+            </div>
+        </div>
+        <div class="leftcolumn">
+            <div class="card">
+                <h2>Hello</h2>
+                <img class="img" src="/images/selfridges2.jpg" alt="shop"></div>
```
amcraethanan: Try to remember to keep the indentation consistent for elements that are nested -  so the div at the end of this line would be on the next line so that its clear where a html element begins and ends. Have another look at this section - there is an extra `</div>` element here.

index.html
```diff
@@ -7,10 +7,92 @@
...
+  <div class="footer">
+    <div class="topnav">
+        <a href="#">Site Map</a>
+        <a href="#">Delivery</a>
+        <a href="#">Contact</a>
+        <a href="#">About Us</a>
+        <a href="#">Privacy Policy</a>
+      </div>
+      <div class="footercol">
+        <div class="footerimagecontainer">
```
amcraethanan: This div could probably be removed if it's not being used for anything.

style.css
```diff
@@ -1 +1,199 @@
...
+  }
+
+  h2 {
+      color:#b81c4c
+  }
+
+  main{
+      max-width: 900px;
+      margin:auto;
+      box-shadow: 30px 0px 40px rgba(0,0,0,0.4), -30px, 0px, 40px rgba(0,0,0,0.4); /*not working*/
```
amcraethanan: This can only be one value, so just:
```
box-shadow: 30px 0px 40px rgba(0,0,0,0.4)
```

This should then add a shadow effect against your whole main element

style.css
```diff
@@ -1 +1,199 @@
...
+  }
+  /* Style the top navigation bar */
+  .topnav {
+    overflow: hidden;
+    margin-top: 10;
```
amcraethanan: reminder to add units so it's clear what it is eg. `10px`

style.css
```diff
@@ -1 +1,199 @@
...
+  
+  /* Style the topnav links */
+  .topnav a {
+    display: block;
+    background-color: rgb(233, 159, 196);
+    color: #850a2f;
+    text-align: center;
+    padding: 14px 16px;
+    text-decoration: none;
+    float: none;
```
amcraethanan: instead of float, use flexbox to use the skills picked from last week 💭

style.css
```diff
@@ -1 +1,199 @@
...
+    width: 100%;
+    
+  }
+  
+  /* Change color on hover */
+  .topnav a:hover {
```
amcraethanan: good job on remembering to use pseudo classes!

style.css
```diff
@@ -1 +1,199 @@
...
+    background-color: #b81c4c;
+    color: white;
+  }
+  
+  /* Create two equal columns that floats next to each other */
```
amcraethanan: i'm not sure if this comment matches what the styling describes here - if you do put comments make sure it's consistent with the output to avoid confusion when you look back on the code later; if you were to do what the comment describes, you would look at using flexbox on the parent class to try and implement this - which is what your `row` class does 😄

style.css
```diff
@@ -1 +1,199 @@
...
+  }
+  
+
+   /* Create 4 equal columns that floats next to each other */ 
+   .bottom-img {
+    
+    background-color: #825947;
+    width: 100%;
+    padding: 2px;
+
```
amcraethanan: remember to tidy up your formatting in your css files too

style.css
```diff
@@ -1 +1,199 @@
...
+    .card {
+        background-color: white;
+        padding: auto;
+        margin-top: 10px;
+      }
+
+    .rightcolumn {   
+        width: 70%;
+        background-color: white;
+        align-self: 2;
```
amcraethanan: have a look at what the valid values are for [align-self](https://developer.mozilla.org/en-US/docs/Web/CSS/align-self) - at the moment the css is invalid and would not apply any alignment to this class

style.css
```diff
@@ -1 +1,199 @@
...
+    display: flex;
+    flex-direction: column;
+    float: right;
+    width: 30%;
+}
+
+  /* main */
+    .card {
+        background-color: white;
+        padding: auto;
```
amcraethanan: Have a look at what the valid values are for [padding](https://developer.mozilla.org/en-US/docs/Web/CSS/padding) - at the moment this css is invalid.

style.css
```diff
@@ -1 +1,199 @@
...
+        background-color: #825947;
+        width: 20%;
+        margin:auto;
+        }
+  }
+  
+  /* Responsive layout - when the screen is more than 400px wide, make the navigation links stack on top of each other instead of next to each other */
+  @media screen and (min-width: 400px) {
+    .topnav a {
+      float: left;
```
amcraethanan: remember to look at using flexbox here instead of float :)

index.html
```diff
@@ -7,10 +7,91 @@
...
+    <div class="upper">
+        <div class="container">
+            <img class="logo" src="/images/logo.png" alt="logo">
+        </div>
+        <div class="container-right">
+            <h1>Your Cake Mate</h1>
+            <h2>For Your Orders <br> 07403720162</h2>
+        </div>
+    </div>
+    <div>
```
LucyMac: You don't need this div here, it doesn't add anything

index.html
```diff
@@ -7,10 +7,91 @@
...
+              <a href="#">About</a>
+            </div>
+        </nav> 
+    </div>
+</header>
+
+
+
+    <div class="row">
+        <div class="rightcolumn">
```
LucyMac: be careful with this kind of class name. On mobile, this is not on the right! Be a little more descriptive of what the section _does_ rather than what it looks like.

style.css
```diff
@@ -1 +1,207 @@
...
+    width: 100%;
+    padding: 2px;
+  }
+
+  /* Footer */
+  .footer {
+
+    width: 100%;
+    float: left;
+    clear: both;
```
LucyMac: Don't use these rules. Flexbox should give you everything you need.

PR #13, opened 2020-02-07
Alexandru

AlexandruBudaca:


style.css
```diff
@@ -1 +1,339 @@
...
+        
+        
+    }
+    .header li {
+        display: inline-block;
+    }
+    .fa {
+        visibility: hidden;
+    }
+    #about {
```
amcraethanan: try not to duplicate classes within the css classes - this already exists below

PR #15, opened 2020-02-07
Responsive cake webpage nader

naderakhgari:


index.html
```diff
@@ -7,10 +7,56 @@
...
+        @import url('https://fonts.googleapis.com/css?family=Open+Sans+Condensed:300&display=swap');
+    </style>
+
+
+
 </head>
 
 <body>
     <!-- Add your markup here -->
+    <nav id="navigation">
```
amcraethanan: Think about using the `<header>` element that was covered in last week's lesson - the nav would semantically make sense to live there in this case.

style.css
```diff
@@ -1 +1,217 @@
...
+    height: 60px;
+}
+#logo{
+    margin-top: 0px;
+    padding-top: 0px;
+
+    height: 50px;
+    width: 120px;
+}
+a{
```
amcraethanan: what happens when you hover over hyperlinks? consider using a pseudo class to style `a` html elements as well

style.css
```diff
@@ -1 +1,217 @@
...
+    margin-top: 10px;
+    margin-left: 0px;
+    width: 480px;
+    
+}
+#second-section{
+    /* height: 235px; */
+    width: 100%;
+    background-color:#f5f6fa;
+    display: flex;
```
amcraethanan: good job on remembering to use flexbox from last week's homework!

style.css
```diff
@@ -1 +1,217 @@
...
+    flex-direction:column;
+    /* position: relative; */
+    margin-top: 10px;
+    align-items: center;
+}
+
+.second-section-img{
```
amcraethanan: Continue to favor using classes where possible in order to encourage re-usability of styles like this. Applying css using id selectors is mainly used when wanting to override class styles.

style.css
```diff
@@ -1 +1,217 @@
...
+        margin-top: 0px;
+    }
+    #logo{
+        margin-top: 0px;
+        padding-top: 0px;
+        height: 90px;
+        width: 200px;
+    }
+    a{
+        text-decoration:none;
```
amcraethanan: if the styles are going to be the same regardless of screen size then they don't need to be written out again within the media query

index.html
```diff
@@ -7,10 +7,58 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Responsive Cake webpage</title>
     <!-- Add a link to your css file here -->
+    <link rel="stylesheet" href="style.css">
+    <style>
+        @import url('https://fonts.googleapis.com/css?family=Anonymous+Pro|PT+Serif&display=swap');
+    </style>
+    <style>
+        @import url('https://fonts.googleapis.com/css?family=Open+Sans+Condensed:300&display=swap');
+    </style>
```
jamesgorrie: _Minor:_ You could import these in the same style tag:
```
 <style>
    @import url('https://fonts.googleapis.com/css?family=Anonymous+Pro|PT+Serif&display=swap');
    @import url('https://fonts.googleapis.com/css?family=Open+Sans+Condensed:300&display=swap');
</style>
```

style.css
```diff
@@ -1 +1,217 @@
 /*  Add your styling here */
+
+#navigation{
```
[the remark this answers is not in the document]
jamesgorrie: What might be better is to use a classname to keep specificity low. [There is a great article here about it](https://css-tricks.com/strategies-keeping-css-specificity-low/).

style.css
```diff
@@ -1 +1,218 @@
...
+    background-color:rgb(214, 235,215);
+    margin-top: 50px;
+    
+}
+#first-text{
+    margin-top: 0px;
+}
+.title{
+    font-family: 'Anonymous Pro', monospace;
+    font-family: 'PT Serif', serif;
```
jamesgorrie: Is there a reason for two `font-family` properties?

style.css
```diff
@@ -1 +1,218 @@
...
+    text-align: center
+            
+}
+.text{
+    font-family: 'Open Sans Condensed', sans-serif;
+    font-size: 25px;
+    color: grey;
+    margin-top: 0px;
+    padding: 5px;
+    margin: 0px 30px 0px 30px;
```
jamesgorrie: _Minor:_ you can use the shorthand `margin: 0 30px`.

style.css
jamesgorrie: [I also find this trick quite helpful](https://github.com/CodeYourFuture/responsive-cake-webpage/pull/6#discussion_r378096499), although it is completely personal preference.

PR #17, opened 2020-02-08
Responsive-cake-website-wireframe/tayfuntayfun

tayfuntayfun:


tayfuntayfun:
This is the mobile view to be converted onto the desktop . Keeping to work on this @media tomorrow.

index.html
```diff
@@ -6,10 +6,70 @@
...
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Responsive Cake webpage</title>
-    <!-- Add a link to your css file here -->
-</head>
-
+    <link rel="stylesheet" href="style.css">
 <body>
+    <header class="mainheader">
+        <div class="topheader">
+            <h3 class="topheader1">LOGO</h3>
```
LucyMac: Logo will probably need to be an `img` tag not an h3

index.html
```diff
@@ -6,10 +6,70 @@
...
+    </section>
+    <section class="cake-wrapper-row">
+        <div class="cake-wrapper-row-part">
+            <figure class="wrapperimage">
+                <div class="cakeswrapper">CAKE1</div>
+                <div class="cakesdetails">Lorem ipsum dolor sit amet, consectetur adipiscing elit</div> 
+            </figure>
+            <figure class="wrapperimage">
+                <div class="cakeswrapper">CAKE2</div>
+                <div class="cakesdetails">Lorem ipsum dolor sit amet, consectetur adipiscing elit</div>
```
LucyMac: You could use `<figcaption>` for this since you have nested it inside a `figure`

style.css
```diff
@@ -1 +1,268 @@
 /*  Add your styling here */
+body{
```
LucyMac: You have 2 files with the same name `style.css`? If you want to split your CSS, make sure you use different file names. If you only need one, then delete the other one 🙂

index.html
tayfuntayfun: REplaced with img tag

style.css
tayfuntayfun: Deleted the redundant one :)

index.html
tayfuntayfun: Changed to figcaption. Learned another thing, thanks Lucy. You are a star.

PR #18, opened 2020-02-08
Responsive cake webpage - Melanie T

melanieet:
Basic outline of my website. I still need to add media queries and more styling.


theseafaringturtle:
Nice work and well organised HTML. A couple of points for improvement:
- The nav header overflows the page on some of the smaller screens
- It is best to use class selectors rather than tag selectors

style.css
```diff
@@ -1 +1,69 @@
...
+
+#header-image {
+    height: 150px;
+    width: 150px;
+    display: block;
+    margin: 0 auto;
+}
+
+.navigation li {
+    padding-right: 50px;
```
LucyMac: This is causing an issue on mobile. Can you think of a way to fix it?

style.css
```diff
@@ -1 +1,69 @@
...
+/* About section styles */
+
+.about {
+    padding: 100px;
+}
+
+/* Example cake styles */
+
+.example-cakes ul {
+   justify-content: space-around;
```
LucyMac: Watch out for default styling on `ul` - inspect this element and you'll see some padding on the left which is setting everything off center. This is a default rule that the browser adds - you can overwrite it with `padding-left: 0`

style.css
```diff
@@ -1 +1,69 @@
...
+
+* {
+    font-family: 'Pacifico', cursive;
+    background-color: rgb(188, 243, 226);
+}
+
+ul {
+    display: flex;
+    justify-content: center;
+}
```
LucyMac: Be careful when styling a generic HTML tag with a specific rule such as `justify-content: center;`
You may well have instance of lists in your site which you don't want to style in the same way. To avoid this, add a class to the element and style that instead.

index.html
```diff
@@ -6,11 +6,56 @@
...
+
+    <main>
+        
+        <section class="about">
+            <h1>Let Them Eat Cake</h1>
+
+            <h3>Feeding the cake deprived since 2020</h3>
+        </section>
+
+        <section class="image-carousel">
```
LucyMac: Keep going @melanieet 😉 
You don't need to make a real carousel but you can fake it with a large image and some smaller 'thumbnails' underneath

melanieet:
Thank you so much for your constructive and encouraging feedback, @theseafaringturtle and @LucyMac  :). 

There's a quite a bit to work on, but I'm enjoying the process, and with your help I know I can improve.

PR #19, opened 2020-02-08
Cake/sabia

sabia-m:
Cake Website HW


style.css
```diff
@@ -61,10 +64,27 @@ h3 {
     padding-left: 30%;
 }
 
-p {
+.display.info {
```
theseafaringturtle: small typo (class name is display-info)

index.html
```diff
@@ -6,11 +6,47 @@
...
+    <link rel="stylesheet" href="style.css">
 </head>
 
 <body>
-    <!-- Add your markup here -->
+<header>
+    <img class="logo" src="images/logo.jpg" alt="logo">
+    <nav>
+        <label for="hamburger">&#9776;</label>
+        <input type="checkbox" class="hamburger"/>
```
theseafaringturtle: If you change hamburger to be an id instead of a class, the checkbox trick will work properly. By the way, JS is usually more convenient for this.

index.html
```diff
@@ -6,11 +6,47 @@
...
+        <div class="slider">
+        <div><img class="slider-image" src="images/starcake.jpg" alt="star cake" ></div>
+        <div><img class="slider-image" src="images/cheesecake.jpg" alt="cheesecake" ></div>
+        <div><img class="slider-image" src="images/pumpkincake.jpg" alt="halloween cake"></div>
+    </div>
+</section>
+<section class="information">
+    <div class="display">
+        <h3>Display</h3>
+        <img class="inside" src="images/display.jpg" alt="inside cake shop" style="width: 60%;">
```
theseafaringturtle: It's best not to mix inline styles in the HTML

theseafaringturtle:
Nice work overall

style.css
```diff
@@ -1 +1,145 @@
...
+
+.slider {
+    display: block;
+}
+
+.display {
+    max-width: 70%;
+    border: 2px solid white;
+    padding: 10px;
+    padding-left: 30%;
```
LucyMac: Why so much padding? It pushes everything to the side. 
Can you try and rework the layout of this section?

PR #20, opened 2020-02-08
Responsive-cake-webpage/tanianm

tanianm:


style.css
```diff
@@ -1 +1,165 @@
...
+.mini-banner p {
+    text-align: center;
+    font-size: 35px;
+    padding-bottom: 60px;
+    margin-bottom: 50px;
+    font-family: 'Segoe UI', Tahoma, Geneva, Verdana, sans-serif;
+}
+
+.btn {
+    justify-content: center;
```
amcraethanan: for this to take effect, you also need to make this a flex container, so `display: flex`

style.css
```diff
@@ -1 +1,165 @@
...
+    width: 100%;
+    height: auto;
+}
+
+.cake-text {
+    font-size: 20px;
+    text-align: center;
+}
+
+hr {
```
amcraethanan: you have this repeated above, make sure there is only definition of styling - remember css reads from the top-down so only the last style definition for a property will be applied

PR #21, opened 2020-02-08
My first pull request

ellietms:
it still  needs some changes ,I am still working on it.it is not finished.


Cake-css.css
```diff
@@ -0,0 +1,123 @@
...
+    align-content:space-around;
+    flex-wrap: wrap;
+}
+.cakes{
+    display: block;
+    box-sizing: border-box;
+}
+.style_cake{
+    height: 300px;
+    padding: 100px 100px;
```
LucyMac: This large padding is interfering with the layout of the page on mobile (the elements are wider than the size of the screen). If you remove the padding and add some extra styling to the `cakes` class (hint: use flexbox), it should fix this.

Cake-css.css
```diff
@@ -0,0 +1,123 @@
...
+.logo_image{
+    display: flex;
+    justify-content: center;
+    height: 200px;
+    background: url("images/Mynavbarimage.jpeg");
+    background-position: center;
+}
+.nav_bar{
+    background: url("images/nav_back.JPG");
+
```
LucyMac: Try adding rules for `background-repeat` and `background-size` here to make the image cover the whole nav

style.css
```diff
@@ -1 +1,182 @@
...
+    border-style: ridge;
+    box-sizing:border-box;
+}
+
+.big_pic{
+    display: flex;
+    flex-direction: column;
+    justify-content:center;
+    background-position: center;
+    background-repeat: no-repeat;
```
LucyMac: This won't do anything if you don't provide a background-image. You've chosen to use an `img` tag inside, which is fine, but you need to add a CSS rule to that `img` so that it is only 100% of its parent.

style.css
```diff
@@ -1 +1,182 @@
...
+}
+
+.width_cakes{
+    width: 50%;
+    height: 50%;
+    text-decoration: none;
+}
+.width_cake3{
+    width: 200px;
+    height: 1%;
```
LucyMac: You're setting a width *and* a height here. As we saw in class, you can't set both otherwise it distorts the image. A solution would be to set `width: 100%` on the 3 images (using the *same* class, not different ones), and then give each `li` (=the parent of the `img` tags) a width (in this case, `width: 33%` would work well because there's 3 of them).

style.css
ellietms: thank you lucy,I dont understand which rule i should use to make my image as parent

PR #22, opened 2020-02-08
Yusuf responsive cake webpage

yusuf963:
this is what I have made so far


yusuf963:
alarm

index.html
```diff
@@ -5,12 +5,85 @@
...
+    <link href="https://fonts.googleapis.com/css?family=Bangers|Caveat|Courgette|Lobster&display=swap" rel="stylesheet">
     <title>Responsive Cake webpage</title>
+    <link href="style.css" rel="stylesheet" type="text/css">
     <!-- Add a link to your css file here -->
 </head>
 
 <body>
-    <!-- Add your markup here -->
+ <main>
+   <header>
```
amcraethanan: Good job for remembering to use the header and main elements, however the header html element should live outside of the main eg.
```
<header>
    ...
</header>
<main>
    ...
</main>
<footer>
    ....
</footer>
```

index.html
```diff
@@ -5,12 +5,85 @@
...
+    <div class="Logo">
+        <img class="logoup" src="Yusuf Logo.png" alt="logo"/>
+        <p>best home made cake delivered at your door</p>
```
amcraethanan: As this forms part of your header, would this make more sense to be a `<h1>` (or a variation of this) element? Alternatively if this text is meant to always be with your img element, consider instead using `<figure>` and `<figcaption>`

index.html
```diff
@@ -5,12 +5,85 @@
...
+                    <li><a href="#">Lesson</a></li>
+                    <li><a href="#">Ordering</a></li>
+                    <li><a href="#">Download recipes</a></li>
+                    <li><a href="#">About</a></li>
+                </ul>
+            </li>
+        </ul>
+    </div>
+  </header>
+<table>
```
amcraethanan: its a nice idea using table however it does mean the when it comes to scaling to smaller screen widths, the table won't be as responsive. Using flexbox container with child elements means your images can then flow without additional styling to smaller screen widths.

index.html
```diff
@@ -5,12 +5,85 @@
...
+        <td><a href="images/cake3.jpg" target="_blank"><img src="images/cake3.jpg"></td>
+    </tr>
+    <tr>
+        <td><a href="images/cake4.jpg" target="_blank"><img src="images/cake4.jpg"></a></td>
+        <td><a href="images/cake5.jpg" target="_blank"><img src="images/cake5.jpg"></a></td>
+        <td><a href="images/cake6.jpg" target="_blank"><img src="images/cake6.jpg"></a></td>
+    </tr>
+</table>
+<p>Our Little Story</p>
+<p>As well as our signature Party Cap Cakes hand baked by us in an ever changing selection of the latest and greatest flavours know to man and woman! we do lots of other tasty stuff too....<br>
```
amcraethanan: Remember to keep an eye on correct indentation...

index.html
```diff
@@ -5,12 +5,85 @@
...
+
+We can talk a long time about the things we love about Cake Stories, such as the beautiful coffee, the huge selection of loose leaf tea, the fact all our food, scones and cakes are made by us, our super friendly and brilliant team, the good hearty freshly prepared food or the many, many lights but the best thing we can do is to invite you for a visit... the door is open!</p>
+<p>See us on Channel</p>
+<div class="iframe-video">
+    
+    <iframe width="560" height="315" src="https://www.youtube.com/embed/hZux_BvIA08" frameborder="0" allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
+</div>
+
+ </main>
+ <div>"
```
amcraethanan: You shouldn't have any elements that don't live in either your header, main or footer elements.

index.html
```diff
@@ -5,12 +5,85 @@
...
+                </li>
+                <li class="facebook">
+                    <i class="fa fa-facebook"></i>
+                 </li>
+                 <li class="instagram">
+                    <i class="fa fa-instagram"></i>
+                 </li>
+           </ul>
+       </div>
+       </div>
```
amcraethanan: This looks like an extra div here...

style.css
```diff
@@ -1 +1,190 @@
...
+    display: flex;
+    justify-content: center;
+    margin: auto;
+    width: 30%;
+}  
+
+/* cart icon */
+h3{
+    color: #ffffff;
+    float: right;
```
amcraethanan: if you want to align text to the right, you can use `text-align` or otherwise use justify-content if this was flexbox

style.css
```diff
@@ -1 +1,190 @@
...
+    background-color: rgb(17, 18, 19, 0.5);
+    padding: 0px 20px 0px 20px;
+}
+ul li{
+    float: left;
+    list-style: none;
+    position: relative;
+
+}
+a{
```
amcraethanan: be careful with your css formatting , this should be `a {`

style.css
```diff
@@ -1 +1,190 @@
...
+   display: block; 
+   text-decoration: none;
+   color:white;
+   font-size: 22px;
+}
+ul li ul{
+    display: none;
+    position: absolute;
+    background-color:rgb(63, 8, 8,.8);
```
amcraethanan: same here, this should be `background-color: rgb(63, 8, 8,.8);`

style.css
```diff
@@ -1 +1,190 @@
...
+    opacity: px;
```
amcraethanan: this is currently invalid css, opacity should always be a number between 0 an 1 and doesn't have a unit eg `opacity: 0.3`

style.css
```diff
@@ -1 +1,190 @@
...
+ul li ul li{
+    width: 120px;
+}
+ul li ul li a{
+    padding: 15px 0px;
+}
+ul li ul li a:hover{
+   background-color: rgb(10, 10, 10, .7); 
+}
+body{
```
amcraethanan: be careful, there are now two instances of `body` being defined in this stylesheet - there should only be one. At the moment, the latest style definition for this html element will apply.

style.css
```diff
@@ -1 +1,190 @@
...
+}
+
+
+/* MEDIA QUARIES */
+
+@media (min-width: 768px){
+    .logoup{
+        display: flex;
+        justify-content: center;
+        margin: auto;
```
amcraethanan: for media queries, you only have to state the properties and html/classes that will be overidden, eg. so these three do not have to be repeated as they're already the same in your general css

index.html
```diff
@@ -5,12 +5,85 @@
...
+   <header>
+    <div class="Logo">
+        <img class="logoup" src="Yusuf Logo.png" alt="logo"/>
+        <p>best home made cake delivered at your door</p>
+     </div>
+     <h3>Add To Cart</h3>
+     <div>
+        <ul>
+            <li><a class="first-menu" href="#">Menu</a>
+                <ul>
```
amcraethanan: Try to avoid having nested `<ul> elements as they can easily get confusing and can be replaced by just using a div with a child ul element

PR #24, opened 2020-02-09
Cake/webpage/abbas

abbasmokhtari:
I set 500px as the breaking point and changed colours and lay out

### DIFF
--- a/index.html
+++ b/index.html
@@ -12,6 +12,7 @@
   <body>
     <!-- Add your markup here -->
 
+    
     <!-- banner and menus -->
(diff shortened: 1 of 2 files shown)


ChrisOwen101:
Hi @abbasmokhtari 

It seems like your styles aren't being used on your website. This is what I see when I open it.
image

Please send me a message if you're still working on this and having trouble. It seems like you've still got some work to do on this homework.

It's never wrong to ask for help, it's only wrong to not ask, The only way that we can help is if you get in contact with us! 🙂

PR #26, opened 2020-02-09
Kyaw RP cake web page

kiowd:
I wrote html and css for the mobile design. But I haven't included images yet but only a few line of text  shown in the example wireframe.

### DIFF
--- a/style.css
+++ b/style.css
@@ -1,366 +1,379 @@
 /*  Add your styling here */
-@import
-  url('https://fonts.googleapis.com/css?family=Crimson+Text&display=swap');
+@import url('https://fonts.googleapis.com/css?family=Crimson+Text&display=swap');
+
 html,
 body {
-  margin: 0;
-  padding: 0;
-  font-family: handlee, sans-serif;
-  box-sizing: border-box;
+    margin: 0;
+    padding: 0;
+    font-family: handlee, sans-serif;
+    box-sizing: border-box;
 }
 
 /* [ON THE SMALL SCREEN] */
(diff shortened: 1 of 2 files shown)


style.css
```diff
@@ -1 +1,383 @@
...
+    /* Show hamburger icon */
+    #page-nav-wrap label {
+        display: inline-block;
+    }
+
+    /* Collapse menu */
+    #page-nav {
+        flex-direction: column;
+        align-items: flex-start;
+        /*align-content: flex-start;*/
```
ChrisOwen101: It's best to remove comments when you're done using them

style.css
```diff
@@ -1 +1,383 @@
...
+    /* Show/hide menu on checkbox click */
+    #page-nav {
+        display: none;
+    }
+
+    #page-nav a {
+
+        border-top: 4px double tomato;
+        width: 100%;
+
```
ChrisOwen101: Remove these empty lines, to tidy up this file a little

style.css
```diff
@@ -1 +1,383 @@
...
+        display: flex;
+        justify-content: space-between;
+        align-items: center;
+        background-color: gainsboro;
+        width: 100%;
+        height: 15vh;
+    }
+
+    .footer a,
+    #page-nav a {
```
ChrisOwen101: Move this onto the line above so it's all on the same line

ChrisOwen101:
This is some outstanding work @kiowd ! If you could get the images in and tidy up the code a little this would be absolutely perfect.

Keep up the great work!

PR #27, opened 2020-02-09
Cake/sonjide

Sonjide:
uploaded all images and working on carousel and media queries - 50% complete


LucyMac:
What have you used in your code @Sonjide? There are so many HTML and CSS files! I can't see which one to look at 😞 
Can you remove any unnecessary files?

PR #29, opened 2020-02-09
Mediaquery/responsive cake design/jreo

jreo123:


index.html
```diff
@@ -6,11 +6,67 @@
...
-    <!-- Add a link to your css file here -->
+    <link rel="stylesheet" href="/style.css">
+
 </head>
 
 <body>
-    <!-- Add your markup here -->
+    
+    <main>
+        <nav>
```
amcraethanan: Good job on remembering to use semantic element `main` for your content however your `nav` should live in the `header` as it is always part of the top of your page.

index.html
```diff
@@ -6,11 +6,67 @@
...
+                <div class="pic1">PIC1</div>
+                <div class="pic2">PIC2</div>
+                <div class="pic3">PIC3</div>
+                <div class="pic4">PIC4</div>
+            </div>
+            
+            <section class="hello">
+            <h1>HELLO</h1>
+        </section>
+        </section>
```
amcraethanan: Remember to keep an eye on your formatting, the indentation on the page should show how nested an element is in orded to clearly see what html element is a child of another.

style.css
```diff
@@ -1 +1,156 @@
-/*  Add your styling here */
+.logo{
+    background-color: hotpink;
+    height: 50px;
+}
+
+
+nav {
+    display: flex;
```
amcraethanan: good job on implementing flexbox :)

style.css
```diff
@@ -1 +1,156 @@
...
+    width: 80px;
+}
+
+.pic2{
+    background-color:yellow;
+    width: 80px;
+}
+.pic3{
+    background-color: lightblue;
+    width: 80px;
```
amcraethanan: instead of having a fixed width for your images, instead use relative sizes like percenatges so that the elements with this class can then scale responsively.

PR #30, opened 2020-02-09
Alisina-cakesite

Alisina92:


style.css
```diff
@@ -1 +1,534 @@
...
+}
+.main-container {
+  background-color: black;
+  color: red;
+}
+.sidebar {
+  width: 25%;
+  display: flex;
+  flex-direction: column;
+  float: left;
```
LucyMac: please don't use `float: left / right`. Always use flexbox instead.

style.css
```diff
@@ -1 +1,534 @@
...
+}
+
+.main-image {
+  width: 75%;
+  float: right;
```
LucyMac: no floats

index.html
```diff
@@ -1,16 +1,101 @@
...
+        <h1 class="title">Hello</h1>
+        <p class="info">information</p>
+      </aside>
 
-<body>
-    <!-- Add your markup here -->
-</body>
+      <section class="main-img-container">
+        <img class="main-image" src="images/vw beetle.jpg" alt="vw-beetle" />
+      </section>
```
LucyMac: You should wrap these 2 elements (sidebar and main-img-container) inside flex container so that the sidebar and the main image appear in a row without using `float: left / right`
Add a section around these and set it to display: flex

style.css
```diff
@@ -1 +1,534 @@
...
+}
+#image-set1-3:target ~ .image-carosel {
+  left: -1600px;
+}
+#image-set1-4:target ~ .image-carosel {
+  left: -2250px;
+}
+.section-container {
+  width: 0;
+}
```
LucyMac: What is this? 
This is causing issues in your layout, please remove it.

style.css
```diff
@@ -1 +1,534 @@
...
+.image-tag2 {
+  width: 100%;
+  display: flex;
+  list-style-type: none;
+}
+.image-set2 {
+  width: 250px;
```
LucyMac: There are 4 images at 250px each, so all together they require at least 1000px wide. This is breaking the layout on small desktop.
A possible solution is to set each `li` to 25%, and the images inside to 100%.

style.css
```diff
@@ -1 +1,534 @@
...
+
+.footer-list {
+  display: flex;
+  list-style-type: none;
+  justify-content: space-around;
+  padding-left: 0%;
+  list-style-type: none;
+  text-align: center;
+  margin: 0;
+  padding: 0;
```
LucyMac: Remove the duplicate rule `padding-left: 0%;` on line 127

index.html
```diff
@@ -1,16 +1,101 @@
...
+      <section class="button-container">
+        <a href="#image-set1-1" class="image-button"></a>
+        <a href="#image-set1-2" class="image-button" id='image2'></a>
+        <a href="#image-set1-3" class="image-button"></a>
+        <a href="#image-set1-4" class="image-button"></a>
+      </section>
+      <section class="section-container">
+        <ul class="image-tag2">
+          <li>
```
LucyMac: If you're using a list and list items, use a list item for each image. 
At the moment you only have one item in your list.

Alisina92:
Hi, lucy I have taken note of your review and will apply accordingly. thanks 
I also would like to have a slack call with you if possible to discuss about cake site

PR #31, opened 2020-02-12
first push

leealex88:
html

### DIFF
--- a/index.html
+++ b/index.html
@@ -55,25 +55,27 @@
       </section>
     </section>
 
-
-    <section>
-      <div>
-        <p class="title">Delicious Works of Art</p>
-        <h6 class="text">by Ola</h6>
+    <section class="pictures-section">
+      <div class="each-picture">
+        <h3>Description</h3>
       </div>
-      <div class="images-cake">
-        <div class="cakes"></div>
-        <div class="cakes"></div>
-        <div class="cakes"></div>
+      <div class="each-picture">
+        <h3>Description</h3>
+      </div>
+      <div class="each-picture">
+        <h3>Description</h3>
+      </div>
+      <div class="each-picture">
+        <h3>Description</h3>
       </div>
     </section>
-    <section id="second-section"></section>
-    <footer id="footer">
-      <h5>© 2020 Ola</h5>
-      <a href="">Contact</a>
-      <a href="">About Us</a>
-      <div></div>
-    </footer>
+
+    <section>
+      <footer class="footer">
+        <a href="">Sitemap</a>
+        <a href="">Contact Us</a>
+        <a href="">Delivery</a>
+      </footer>
   </main>
 </body>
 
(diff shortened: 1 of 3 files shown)


leealex88:
Hi, I did finish my homework late, apologise for this. I was trying to change the height of my .first-section and .second-section for 300px when the user is viewing it on a desktop, but when I am adding a heigh in a @media query than it does not work.

index.html
```diff
@@ -2,15 +2,81 @@
...
 
 <body>
-    <!-- Add your markup here -->
+  <main>
+    <header>
+      <nav>
+        <div class="grid-one">
+          <img class="logo"
+            src="https://images.pexels.com/photos/1337380/pexels-photo-1337380.jpeg?auto=compress&cs=tinysrgb&dpr=2&w=500"
+            heigh="120" width="120" />
```
LucyMac: Your logo probably doesn't need to be part of the `nav`

index.html
```diff
@@ -2,15 +2,81 @@
...
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+  <title>Responsive Cake webpage</title>
+  <link rel="stylesheet" href="style.css" />
 </head>
 
 <body>
-    <!-- Add your markup here -->
+  <main>
+    <header>
```
LucyMac: Your header should be outside of `main`.

style.css
```diff
@@ -1 +1,126 @@
...
+    }
+
+    .grid-two {
+        display: flex;
+        justify-content: space-around;
+    }
+
+    .general-hero-section {
+        display: flex;
+        padding: 10px;
```
LucyMac: You might want all your sections to line up on the edges of the screen. This section seems a little more indented on desktop than the others, and your `pictures-section` also has a different spacing. Try and stay consistent for your design to look nice and neat 🙂

style.css
```diff
@@ -1 +1,126 @@
...
+    .each-picture {
+        margin: 5px;
+    }
+
+    .footer {
+        justify-content: flex-start;
+    }
+}
+
+.grid-two {
```
LucyMac: Watch out with class names like this - it implies that it does something to the layout but in reality it doesn't.

index.html
```diff
@@ -2,15 +2,81 @@
...
+      </div>
+      <div class="each-picture">
+        <h3>Description</h3>
+      </div>
+      <div class="each-picture">
+        <h3>Description</h3>
+      </div>
+    </section>
+
+    <section>
```
LucyMac: You don't need to wrap your footer in a section. `footer` is enough to tell the browser that this is a new section _and_ that it's a footer.
Also, move the footer outside of `main`. Your structure should be:
```
<header></header>
<main></main>
<footer></footer>
```

leealex88:
Lucy, thank you so much for this very useful comments! I implemented everything and it helps me to remember it better, for the future. Please keep doing it like this! thank you <3

PR #32, opened 2020-02-12
Class 6 cake/tam

TamBahri:


index.html
```diff
@@ -1,16 +1,105 @@
...
+        </div>
+          <div class="btn">
+              <button>
+                Contact
+
+              </button>
+            
+            </div>
+    </header>
+    <section>
```
LucyMac: You probably don't need this section here as it just wraps the nav and doesn't add any styling

index.html
```diff
@@ -1,16 +1,105 @@
...
+                
+
+        </nav>
+    </section>
+    <section class="background">
+
+    </section>
+    <section class="container">
+        <div class="row">
+            <div class="col-12 md-col-12">
```
LucyMac: Great that you're using the grid system!! ⭐️
Just so you know, because this is a 'mobile-first' grid system, `col-12` will work from a tiny screen all the way up to a large screen. You only need to use `md-col-...` if you are _changing_ the number of columns (for example `col-12 md-col-6`). 
What you've done won't cause any problems, but you could just say `col-12` and it would be the same :)

style.css
```diff
@@ -1 +1,148 @@
...
+    justify-content: space-around;
+    padding: 0 10px;
+   
+}
+.col-12{
+    width: 100%;
+    padding: 20px 10px;
+    margin-top: 10px;
+
+}
```
LucyMac: Don't write your own grid - copy and paste the code from here instead: https://github.com/CodeYourFuture/simple-grid/blob/master/grid.css

And *don't change it*. Remove all your own `col-...` classes.

style.css
```diff
@@ -1 +1,148 @@
...
+.col-12 img{
+    width: 100%;
+}
+.col-6{
+    width: 35%;
```
LucyMac: 6 columns is half of 12 columns, so it's *always* 50%.

Please remove all these and replace with the pre-written grid linked above.

style.css
```diff
@@ -1 +1,148 @@
...
+@media (min-width: 700px) {
+    .md-col-3{
+
+    width: 22%;
+
+    }
+    .md-col-12{
+        width: 40%;
+    }
+}
```
LucyMac: Remove these as well once you've added the other grid. Otherwise this CSS will clash with the grid.

index.html
```diff
@@ -1,16 +1,105 @@
...
+            <div class="col-6 md-col-3">
+                Cupcake
+             <img class="product" src="https://cdn.shopify.com/s/files/1/1703/7103/products/Valentines_2019_box_12_minis_rev0_1000x.jpg?v=1544194204" alt="cup cake">
+            </div>
+        </div>
+     </section>  
+    
 </body>
+<footer class="footer">
+    <div class="content">
```
LucyMac: You probably don't need this. It wraps the same elements as the footer, and the `content` class doesn't do anything in your CSS.

PR #38, opened 2020-06-02
Daniel carter week 3

carterd888:
so far designed page for around 480px (mobile view)

[using this wireframe](https://wireframe.cc/Uu79e4)

cake on mobile


carterd888:
have updated pull request with new commits. any feedback would be great, have kept the coloured backgrounds i used for breakpoints as a visual aid.

PR #40, opened 2020-06-04
html-css/week3/Gintaras

g1st:


style.css
```diff
@@ -1 +1,464 @@
...
+.example-cakes-wrapper {
+  margin-top: 7rem;
+}
+
+.example-image {
+  max-width: 100%;
+  display: block;
+  margin: 2rem auto;
+  border-radius: 6px;
+  box-shadow: 0px 6px 6px 0px rgba(0, 0, 0, 0.4)
```
pddys, line 230: Because of the hover scale effect, you could add a transition here to make the change less abrupt.

PR #62, opened 2020-11-24
Birmingham Class 2 - Chizim Chinuru - HTML/CSS - Week_3

scarabeo7:
- Your Name: Chizim Chinuru
- Your City: Birmingham
- Your Slack Name: Chizim

# Homework Details

- Module: HTML-CSS-Coursework
- Week: Week-3


index.html
```diff
@@ -1,16 +1,124 @@
...
+                    <a id="footer-links" href="#">Delivery</a>
+                </li>
+               <li>
+                    <a id="footer-links" href="#">Contact</a>
+                </li>
+           </ul>
+       </footer>
+
+
+    <script src="https://code.jquery.com/jquery-3.5.1.slim.min.js" integrity="sha384-DfXdz2htPH0lsSSs5nCTpuj/zy4C+OGpamoFVy38MVBnE+IbbVYUew+OrCXaRkfj"
```
ARTaylor1502, line 118: CDN declarations should go in the head

scarabeo7: I could be very wrong on this but we were told to add script just under the body tag? Is that info wrong?

ARTaylor1502: CDNs and libraries should be included in the head as they are classed as dependencies. i.e code below them in the DOM relies on them to work. You should generally put functions and code that directly affects the DOM in the body at the end of the last child element. This is because if the code requires DOM element to reference (which it almost certainly will) it will need to wait for the DOM elements in question to be loaded first and made accessible. i.e if the script is at the start the elements wont be accessible to manipulate yet.

A quick way to think of it is pretty much:

head - libraries, CDNs, functions which need to be called on page load (e.g google analytics scripts)
body - everything else that depends on interacting with the DOM elements

ARTaylor1502: So yes you were taught very much correctly, there are just few exceptions to the rule to be aware of

index.html
```diff
@@ -1,16 +1,135 @@
...
+                            <img src="/images/20200719_164740.jpg" alt="" class="img-fluid img-thumbnail max-width: 100% height:auto">
+                        </div>
+                        <div class="col-sm-2">
+                            <img src="/images/20200805_131129.jpg" alt="" class="img-fluid img-thumbnail max-width: 100% height:auto">
+                        </div>
+                        <div class="col-sm-2">
+                            <img src="images/ulysse-pointcheval-1WmlAiYgnoI-unsplash.jpg" alt="" class="img-fluid img-thumbnail max-width: 100% height:auto">
+                        </div>
+                        <div class="col-sm-2">
+                            <img src="/images/20200706_171336.jpg" alt="" class="img-fluid img-thumbnail max-width: 100% height:auto">
```
ARTaylor1502, line 92: You cannot use these styling properties in a class value. max-width: 100% height:auto won't work. They need to be either declared inline like this: style="max-width: 100% height:auto" or better still in the style.css stylesheet.

ARTaylor1502: Also some of the image src are declared absolute from the route. I would remove the first '/' from each of these source attributes

index.html
```diff
@@ -1,16 +1,135 @@
...
-</body>
+    <head>
+        <meta charset="UTF-8">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <meta http-equiv="X-UA-Compatible" content="ie=edge">
+        <link href="https://fonts.googleapis.com/css?family=Tangerine&amp;display=swap" rel="stylesheet">
+        <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap@4.5.3/dist/css/bootstrap.min.css" 
+            integrity="sha384-TX8t27EcRE3e/ihU7zmQxVncDAy5uIKz4rEkgIXeMed4M0jlfIDPvg6uqKI2xXr2" crossorigin="anonymous">
+        <link rel="stylesheet" href="/css/style.css">
+        <script src="https://code.jquery.com/jquery-3.5.1.slim.min.js" integrity="sha384-DfXdz2htPH0lsSSs5nCTpuj/zy4C+OGpamoFVy38MVBnE+IbbVYUew+OrCXaRkfj"
```
ARTaylor1502, line 12: Not sure jquery is needed?

index.html
```diff
@@ -1,16 +1,135 @@
...
+                            <img src="images/cake3.jpg" alt="" class="img-fluid img-thumbnail" id="images">
+                        </div>
+                        <div class="col-sm-3">
+                            <img src="images/cake4.jpg" alt="" class="img-fluid img-thumbnail" id="images">
+                        </div>
+                        <div class="col-sm-3">
+                            <img src="images/cake5.jpg" alt="" class="img-fluid img-thumbnail" id="images">
+                        </div>
+                    </div>
+
```
ARTaylor1502, line 95: Try to avoid spaces like this, it can make your code look inconsistent.

index.html
```diff
@@ -1,16 +1,135 @@
...
+        <script src="https://cdn.jsdelivr.net/npm/bootstrap@4.5.3/dist/js/bootstrap.bundle.min.js" 
+            integrity="sha384-ho+j7jyWK8fNQe+A12Hb8AhRq26LrZ/JpcUGGOn+Y7RsweNrtN/tE3MoK7ZeZDyx" crossorigin="anonymous"></script> 
+        <title>Responsive Cake webpage</title>       
+    </head>
+
+    <body>
+        <nav class="navbar navbar-expand-lg navbar-dark bg-dark static-top">
+            <div class="container">
+              <a class="navbar-brand" href="#">
+                <img src="images/logo.jpg" alt="Company Logo" id="logo-image">
```
ARTaylor1502, line 23: You need to set a max width for this image. When I checked this branch out locally the nav bar image was enormous and took up the whole of the browser's view port.

index.html
```diff
@@ -1,16 +1,135 @@
...
+                  <li class="nav-item">
+                    <a class="nav-link" href="#">About</a>
+                  </li>
+                  <li class="nav-item">
+                    <a class="nav-link" href="#">Contact</a>
+                  </li>
+                </ul>
+              </div>
+            </div>
+          </nav>
```
ARTaylor1502, line 51: Be careful with your tabbing, this closing nav tab should be level with the start of the main tab underneath it as they are sibling elements.

index.html
```diff
@@ -1,16 +1,135 @@
...
+
+       <main class="container-fluid">
+           <div class="row">
+               <div class="col-sm-4">
+                   <h1 class="introduction text-center" >
```
ARTaylor1502, line 56: Avoid leaving trailing spaces within elements

index.html
```diff
@@ -1,16 +1,135 @@
...
+                        </div>
+                    </div>
+
+              </div>
+
+           </div>
+           
+           <div class="row">
+                <div class="col-sm-3">
+                    <img src="images/cake6.jpg" alt="" class="img-fluid img-thumbnail" id="images">
```
ARTaylor1502, line 102: You have duplicate ids with the name 'images'. This defeats the object of ids as they should be unique. The browser will actually highlight this issue now and can affect your site's google rankings I believe.

index.html
```diff
@@ -1,16 +1,135 @@
...
+                    <figcaption>Some info Cake</figcaption>
+                </div>
+                <div class="col-sm-3">
+                    <img src="images/cake7.jpg" alt="" class="img-fluid img-thumbnail" id="images">
```
ARTaylor1502, line 106: Alt attribute needs to be filled in for accessibility reasons. i.e for those with visual impairments to know what the image represents.

index.html
```diff
@@ -1,16 +1,135 @@
...
-    <!-- Add your markup here -->
-</body>
+    <head>
+        <meta charset="UTF-8">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <meta http-equiv="X-UA-Compatible" content="ie=edge">
+        <link href="https://fonts.googleapis.com/css?family=Tangerine&amp;display=swap" rel="stylesheet">
+        <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap@4.5.3/dist/css/bootstrap.min.css" 
+            integrity="sha384-TX8t27EcRE3e/ihU7zmQxVncDAy5uIKz4rEkgIXeMed4M0jlfIDPvg6uqKI2xXr2" crossorigin="anonymous">
+        <link rel="stylesheet" href="/css/style.css">
```
ARTaylor1502, line 11: The link for this stylesheet is broken when I check this out locally. Should be 'css/style.css' without the preceding '/'.

ARTaylor1502:
Excellent. Looks very close to the layout detailed in the brief's wireframe. Semantic code and good understanding of the bootstrap framework.

scarabeo7:
Thanks for the reviews. I have refactored code as per review if you want to have a look again.

ARTaylor1502:
Just checked the resolutions to the comments. Very good, thanks for taking the time 👍

PR #72, opened 2020-11-27
West Midlands 2 - Haroon Iqbal - HTML/CSS - Week 3

hiqbal0121:
# My Details

- Your Name: Haroon Iqbal
- Your City: Birmingham
- Your Slack Name: hiqbal0121

# Homework Details

- Module: HTML/CSS
- Week: 3


pddys (commented):
Overall this shows that you are comfortable using HTML and CSS. The purpose of this exercise was to design a responsive website using media queries and breakpoints. While some responsiveness has been handled with flex, there is only a single breakpoint on this, so there is not much to judge whether you understand the concept.

The use of semantic html and use alt tags on images is good, and the commit messages are descriptive and well organised.

PR #76, opened 2020-11-27
kossivi/Responsive-Cake-Webpage

Kos0606:
# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

### DIFF
--- a/index.html
+++ b/index.html
@@ -7,10 +7,51 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Responsive Cake webpage</title>
     <!-- Add a link to your css file here -->
+    <link rel="stylesheet" href="style.css">
+
 </head>
 
 <body>
     <!-- Add your markup here -->
+    <div class="flex-container">
+        <div>LOGO</div>
+        <div>CONTACT</div>
+     
+    </div>
+
+    <div class="nav-bar">
+        <div>ABOUT</div>
+        <div>LESSONS</div>
+        <div>ORDERING</div>
+        <div>CAKES</div>
+        <div>HOME</div>
+    </div>
+
+    <div class="main-content">
+        <div>HELLO
+            <p>INFORMATION</p>
+        </div>
+
+        <div>IMAGE CAROUSEL
+            <div>IMG1</div>
+            <div>IMG2</div>
+            <div>IMG3</div>
+            <div>IMG4</div>
+        </div>
+    </div>
+
+    <div class="order-list">
+        <div>CAKE1</div>
+        <div>CAKE2</div>
+        <div>CAKE3</div>
+        <div>CAKE4</div>
+
+    </div>
+<footer class="extra-info">
+    <div>SITE MAP</div>
+    <div>CONTACT</div>
+    <div>DELIVERY</div>
+</footer>
 </body>
 
 </html>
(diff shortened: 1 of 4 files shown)


index.html
```diff
@@ -7,10 +7,70 @@
...
 <body>
     <!-- Add your markup here -->
+    <div class="flex-container">
+        <div>LOGO</div>
+        <div>CONTACT</div>
+     
+
+    </div>
+
+<br>
```
pddys, line 23: While it's valid HTML, generally it's better to use css padding and margin instead of <br> to create spacing.

style.css
```diff
@@ -1 +1,75 @@
...
+
+.flex-container {
+  display: flex;
+  background-color: #d3d3d3;
+  justify-content: space-around;
+}
+
+.flex-container > div {
+  background-color: #a9a9a9;
+  margin: 0.627em;
```
pddys, line 11: There are some values here  without much context as to why they exist. These are called 'magic numbers'.

It may be useful to add a comment explaining why you used these decimals.

style.css
```diff
@@ -1 +1,75 @@
...
+  padding: 1.255em;
+  font-size: 1.882em;
+}
+
+.nav-bar {
+  display: flex;
```
pddys, line 17: Good use of flex.

style.css
```diff
@@ -1 +1,75 @@
...
+  justify-content: space-around;
+  flex-direction: row-reverse;
+}
+
+.nav-bar > div {
+  background-color: #0a0101;
+  margin: 0.627em;
+  padding: 0.1em;
+  font-size: 1.882em;
+  color: white;
```
pddys, line 28: You've mixed hex colors and web safe colours here. It's best to stick to one format.

style.css
```diff
@@ -1 +1,75 @@
...
+}
+
+.main-content > div {
+  background-color: #d3d3d3;
+  margin: 0.627em;
+  padding: 3.137em;
+  font-size: 1.875em;
+}
+
+.main-content div:nth-child(2) > p {
```
pddys, line 44: In this case, it may have been better to add a new class, than target paragraphs using nth-child.

style.css
```diff
@@ -1 +1,75 @@
...
+
+.extra-info > div {
+  background-color: #f1f1f1;
+  margin: 0.627em;
+  padding: 0.937em;
+  font-size: 1.875;
+}
+
+@media only screen and (max-width: 100%) {
+  body {
```
pddys, line 73: Was this left unfinished?

style.css
```diff
@@ -1 +1,75 @@
...
+}
+
+.extra-info > div {
+  background-color: #f1f1f1;
+  margin: 0.627em;
+  padding: 0.937em;
+  font-size: 1.875;
+}
+
+@media only screen and (max-width: 100%) {
```
pddys, line 72: max-width would take in absolute values rather than 100%.

pddys (commented):
This exercise looks unfinished. The instructions were to start from a mobile width and then use media queries to create 3 different versions of the layout. I can only see one layout at desktop widths.

Was there more to this that was not committed? Otherwise this looks like the starting point, rather than a finished assignment.

PR #83, opened 2020-12-13
London Class 7 - Rafael Flores - HTML/CSS - Week 3

rfloresram:
Giving styles to boxes

# Your Details

- Your Name: Rafael Flores
- Your City: London
- Your Slack Name: Rafael F.

# Homework Details

- Module: HTML/CSS
- Week: Week 3


rfloresram (commented):
Please, could I have your feedback? Thanks!

index.html
```diff
@@ -6,11 +6,83 @@
...
+<section class="bakery-icon">
+    <div class="cupcake-icon">
+        <img src="./design/cupcake_logo.png" alt="Bakery allusive icon">
+    </div>
+<br>
+<div>
+    <p class="welcome-msg">The best cakes in town delivered to your door</p>
+</div>
+
+</section>
```
LucyMac, line 27: Watch your indentation please, it makes it hard to read.

index.html
```diff
@@ -6,11 +6,83 @@
...
+          <div class="menu-nav" id="links">
+            
+            <ul>
+                <li><a href="#HOME">HOME</a></li>
+                <li><a href="#CAKES">CAKES</a></li>
+                <li><a href="#ORDERING">ORDERING</a></li>
+                <li><a href="#LESSONS">LESSONS</a></li>
+                 <li><a href="#ABOUT">ABOUT</a></li>
+            </ul>
+        </div>
```
LucyMac, line 45: menu-nav seems to be your desktop nav - so it shouldn't live inside the `mobile-nav` section.

Also, the `manu-nav` `div` doesn't seem necessary - you could move the `menu-nav` class to the `ul` and remove the `div`. We want to avoid using more HTML than we need to.

index.html
```diff
@@ -6,11 +6,83 @@
...
+                <img class="welcome-image" src="./design/bday_cake.jpg" alt="Cake picture">
+            </div>
+        </section>
+        <section class="products">
+                <img src="./design/blue_cupcake.jpg" alt="cupcake w/ blue background">
+                <img src="./design/pink_macaroon.jpg" alt="macaroons w/ pink background">
+                <img src="./design/yellow_cookie.jpg" alt="cookie w/ yellow background">
+                <img src="./design/blue_donut.jpg" alt="donut w/ blue background">
+        </section>
+    </main>
```
LucyMac, line 68: Nice structure 👍 
The `products` section could probably be a list.

index.html
```diff
@@ -6,11 +6,83 @@
...
 </body>
 
+<!-- Footer / Social media / Contact -->
+
+<footer>
+    <section class="social-media">
+        <p>Find us</p>
+    </section>
```
LucyMac, line 76: This doesn't need a separate section - it's just the text above the media-icons, so it can probably be just the `p` by itself

style.css
```diff
@@ -1 +1,217 @@
-/*  Add your styling here */
+/* General body */
+
+body{
+    background-color: rgb(255, 255, 255);
+    font-family: Arial !important;
```
LucyMac, line 5: Never use `!important`!
If you feel that you need it, you are probably trying to override a rule to a selector somewhere that is more 'specific' than `body`
[More details here ](https://dev.to/emmabostian/css-specificity-1kca)

style.css
```diff
@@ -1 +1,217 @@
...
+
+.bakery-icon{
+    width: 100%;
+    display: flex;
+    justify-content: space-around;
+}
+
+.bakery-icon p{
+    font-size: calc(15px + 1vw);
+}
```
LucyMac, line 27: You have a class for this element: `welcome-msg` so it's best to use that in this case.

Good that you've discovered the `calc()` rule 🙂 however `15px + 1vw` seems like a strange combination here. `vw` is relative to the width of the page, which is probably not relevant for a font-size.

style.css
```diff
@@ -1 +1,217 @@
...
+    width: 80%;
+    
+}
+.menu-nav {
+    overflow: hidden;
+    background-color:rgb(207, 144, 155);
+    position: absolute;
+    display:none;
+    min-width: 100vw;
+    top:calc(65px + 3vw);
```
LucyMac, line 63: The `top` property only works in conjunction with `position: absolute` or `position: fixed`

style.css
```diff
@@ -1 +1,217 @@
...
+    justify-content: right;
+    text-align: end;
+    font-weight: bold;
+}
+
+/* Navigation bar */
+
+#links{
+    display:none;
+}
```
LucyMac, line 49: Don't use IDs when you can use a class. Again, this is for specificity reasons.

style.css
```diff
@@ -1 +1,217 @@
...
+
+.welcome-image{
+    width: 100%;
+    height: 100%;
+    flex: auto;
+    margin-bottom: 25px;
+}
+
+.welcome{
+    width: 100;
```
LucyMac, line 90: This is missing a unit 😉

style.css
```diff
@@ -1 +1,217 @@
...
+      margin:0 5px;
+      
+   } 
+
+     /* Main */
+
+    .welcome{
+        display: flex;
+        justify-content: space-between;
+        margin: 2rem 2 rem ;
```
LucyMac, line 191: This rule is not being applied because there is an extra space between `2` and `rem` 🙁 
You can inspect your CSS in the browser's dev tools to see what it actually being applied.

LucyMac (commented):
Nice work Rafael, I love your design and your page is responsive 👍 
Just watch out for your indentation, and also for CSS rules that are incorrect - check your dev tools to understand this!
You could also add an extra media query to improve the layout on tablet.

PR #85, opened 2020-12-13
London Class 7 - Ionut Tripa - HTML/CSS - Week 3

iTrip23:
# Your Details

- Your Name: Ionut Tripa
- Your City: London
- Your Slack Name: Johnny

# Homework Details

- Module: HTML & CSS
- Week: 3


style.css
```diff
@@ -1 +1,267 @@
...
+    font-size: 1.3rem;
+  }
+
+  .cake-pictures {
+    height: 30vh;
+    border-radius: 0.5em;
+  }
+}
+
+@media (min-width: 1000px) {
```
LucyMac, line 193: Well done for using 2 media queries 👌

style.css
```diff
@@ -1 +1,267 @@
...
+  }
+
+  .p-header {
+    font-size: 2rem;
+    letter-spacing: 0.1em;
+  }
+
+  .cake-pictures {
+    width: 20vw;
+    height: 30vh;
```
LucyMac, line 212: Watch out here, don't set a width AND a height to your image, it's causing distortion. You can wrap each image with a div which can have width and height, and then set the `img` to be `width: 100%` but don't set a height. The img will then naturally fill the width of the wrapper div, but without losing its aspect ratio

LucyMac (commented):
Great HTML structure - semantic and organised - and love the site name 😉 
Your CSS is good - great that it uses 2 media queries in a mobile-first way, and I like that you've experimented with hover effects 👍 
Just be careful with the sizing of elements - something is forcing your page to be wider than 100%, which is creating a horizontal scroll. Replacing your margins with padding should improve this.

PR #91, opened 2020-12-13
London class 7/murat demirtas/html css coursework week3

muratdemirtas2001:
# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:


index.html
```diff
@@ -7,10 +7,127 @@
...
+        </section>
+        
+    </main>
+
+    <footer>
+        <section class="footer-heading">
+            <p>Join us on</p>
+        </section>
+        <section class="footer-icons">
+            <img src="./design/instagram-icon.svg">
```
makanti, line 108: Make sure you add alt attributes on all images for accessibility reasons ;)

index.html
```diff
@@ -7,10 +7,127 @@
...
+            <img src="./design/twitter-icon.svg">
```
makanti, line 109: ☝️

index.html
```diff
@@ -7,10 +7,127 @@
...
+            <img src="./design/facebook-icon.svg">
+        </section>
+        <section class="footer-copyright">
+            <p>&copy My cake,Inc.</p>
```
makanti, line 113: Missing `;` from html entity `&copy;` ? 😉

style.css
```diff
@@ -1 +1,270 @@
...
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+}
+
+footer p{
+    font-size: calc(10px + 1vw);
+}
+
+@media(min-width:540px){
```
makanti, line 197: Great use of media queries 👌

style.css
```diff
@@ -1 +1,270 @@
...
+     }
+     nav{
+         width: 100%;
+     }
+     .mobile-screen{
+         width: 100%;
+     }
+     .topnav{
+         width: 100%;
+         display: block;
```
makanti, line 236: This is overwritten by the rule that follows so you can remove it.

index.html
```diff
@@ -7,10 +7,127 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Responsive Cake webpage</title>
     <!-- Add a link to your css file here -->
+    <link rel="stylesheet" href="./style.css">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
```
makanti, line 11: 👏 Also, this may be helpful here: https://www.cdnplanet.com/blog/faster-google-webfonts-preconnect/

style.css
```diff
@@ -1 +1,270 @@
 /*  Add your styling here */
+/* styling header section */
+body{
+    margin:0px;
+    font-family: 'Langar', cursive;
+}
+header{
```
makanti, line 7: Type selector `header` is very generic. `<header>` tag may be used multiple times in an html document (eg inside an `<article>`) which means that the styles you write here will be applied on all headers you may add on the page. Use more specific selectors like class selector to avoid confusion. A class selector will also be arguably more efficient. Check this out: https://csswizardry.com/2011/09/writing-efficient-css-selectors/ 😉

style.css
```diff
@@ -1 +1,270 @@
...
+.icon-img img{
+    width:100%;
+}
+.icon p{
+    font-size: calc(10px + 1vw);
+    width:100%;
+    color:rgb(143, 63, 76);
+}
+
+nav{
```
makanti, line 39: Same applies here (see comment around `header` type selector

style.css
```diff
@@ -1 +1,270 @@
...
+    justify-content:space-around;
+    align-items: center;
+    flex-wrap:nowrap;
+}
+.products{
+    display: flex;
+    flex-direction: column;
+    row-gap:25px;
+}
+.square-cakes, .rounded-cakes{
```
makanti, line 126: It looks like you only need one class for cake boxes as all the styles are shared
How about `.cake-box`. Then you can get rid of the duplicate `.rounded-cakes p` and `.square-cakes p` styles as well

style.css
```diff
@@ -1 +1,270 @@
...
+
+.products img:hover{
+    border:5px solid pink;
+}
+.square-cakes p{
+    text-align:center;
+    font-size: calc(10px + 1vw);
+    color:rgb(143, 63, 76);
+}
+.rounded-cakes p{
```
makanti, line 145: Check previous comment around duplicate styles

style.css
```diff
@@ -1 +1,270 @@
...
+      .icon-text{
+        text-align: right;
+    }
+    .hamburger-icon{
+         display:none;
+     }
+     header{
+         flex-direction: column;
+     }
+     nav{
```
makanti, line 228: Use class selector instead of type selector (more on comment above)

style.css
```diff
@@ -1 +1,270 @@
...
+    .icon-img{
+        z-index: 2;
+      } 
+      .icon-text{
+        text-align: right;
+    }
+    .hamburger-icon{
+         display:none;
+     }
+     header{
```
makanti, line 225: Use class selector instead of type selector (more on comment above)

makanti (commented):
Great effort Murat 👏🥳! Your page looks nice and responsive. You have applied the media queries as expected. You have used semantic markup and you worked well on styling your cakes website.

Try to add some more spacing around your content especially on the desktop version of the page to make it look cleaner.

Also, one thing to improve is the hover effect of your images. You will notice that when you hover over the images, because the border is being added the elements that are below the hovered element are being pushed down. This is because the size of the element changes when you add a border on it. [This](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/The_box_model) will help you understand why.
One way to solve this is by using `outline` instead of `border` property.
On a different note, it is advised that hover state should be mainly applied to clickable elements (eg `<a>` tags) to indicate to the user that they can  click on them. 

I have added a few more comments below but again very good effort overall!

index.html
muratdemirtas2001: thank you Makanti, I added alt attributes to the images

index.html
muratdemirtas2001: Thank you I added it

style.css
muratdemirtas2001: thank you very much

style.css
muratdemirtas2001: thank you I deleted it

style.css
muratdemirtas2001: Thank you I deleted unnecessary classes and combine them

style.css
muratdemirtas2001: I have done it thank you

PR #93, opened 2020-12-15
London Class 7 - Askin Ekinci - HTML/CSS - Week 3

askinekincihf:
# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:


index.html
```diff
@@ -1,16 +1,113 @@
...
+      crossorigin="anonymous"
+    ></script>
     <!-- Add a link to your css file here -->
-</head>
+  </head>
+
+  <body>
+    <header class="header">
+      <div class="header-container">
+        <a class="header-img"><img src="/img/logo.jpg" /></a>
```
makanti, line 19: Remember to always use alt attributes on a tags for accessibility reasons

style.css
```diff
@@ -1 +1,267 @@
...
+    border-top: 1px solid rgb(190, 110, 35);
+    width: 100%;
+}
+
+.hero-desc p {
+    padding: auto 1rem;
+}
+
+.hero img {
+    width: 100%;
```
makanti, line 115: It looks like the images are stretched in certain window sizes. This is caused by forcing a width and height on them which don't obey the image's aspect ratio. The preferable solution to this is to use similar aspect ratio images (eg images with the same width and height) and make sure the width and height you apply respects that ratio.

A different option would be to use div elements instead of img. You could then use css `background` property to apply an image to the div and use `background-size` properties you get a few more options on how to size and position your images.

style.css
```diff
@@ -1 +1,267 @@
...
+}
+
+.company-name {
+    text-align: center;
+    margin-bottom: 1rem;
+}
+
+/* Responsive Rules */
+
+@media (min-width: 600px) and (max-width: 900px) {
```
makanti, line 170: `max-width` is not needed here.
Check `readme.md` for more detail on how to create mobile first styles
Try changing this to `@media (min-width: 600px)` and see what happens 😉

style.css
```diff
@@ -1 +1,267 @@
...
+
+.menu-btn {
+    padding-right: 2rem;
+    padding-top: 1rem;
+    font-size: 2rem;
+    cursor: pointer;
+}
+
+.menu-btn:hover {
+    font-size: 2.5rem;
```
makanti, line 40: Well done for using `rem` 👏

style.css
```diff
@@ -1 +1,267 @@
-/*  Add your styling here */
+/* General rules */
+* {
+    margin: 0;
+    padding: 0;
+    box-sizing: border-box;
+}
+
+body {
+    font-style: italic;
```
makanti, line 9: By adding this rule here, you are making a choice to use italic text globally in your page. This can cause the following issue:
Some of the html tags like `<em>` and `<i>` will be applying italic styling by default when used to add emphasis on the text. This means that if you use italics everywhere on your page you lose the capability of emphasising text in your document, which is bad for accessibility reasons. A way to solve this is to remove the rule from body and apply it to more specific selectors or apply different styles to `em` and `i` selector to differentiate them. Eg `em,i { font-style: bold; }`    See this for more detail: https://css-tricks.com/how-to-italicize-text/

index.html
```diff
@@ -1,16 +1,113 @@
...
+            <li>
+              <a>Contact Us</a>
+            </li>
+            <li>
+              <a>Careers</a>
+            </li>
+          </ul>
+        </div>
+        <!-- Footer Policies -->
+        <div class="class="footer-list-policies">
```
makanti, line 94: `class="footer-list-policies"` ?

style.css
```diff
@@ -1 +1,267 @@
...
+    margin-top: 1rem;
+    padding-top: 0.5rem;
+    text-align: center;
+    margin-bottom: 0.5rem;
+    border-top: 1px solid rgb(190, 110, 35);
+    width: 100%;
+}
+
+.hero-desc p {
+    padding: auto 1rem;
```
makanti, line 111: This is not valid css. Padding cannot be `auto`. [Check this out](https://developer.mozilla.org/en-US/docs/Web/CSS/padding
)

index.html
makanti, line 19: Logo image is missing. Try `src="img/logo.jpg" ` instead :)

makanti (commented):
Well done Askin 🌟👏! You have created a nice responsive cake page, you used semantic markup and made the most of flex. Well done for keeping the code tidy and well formatted

You will notice a few issues with image resizing when moving from one screen size to the other. See comments below on advice around how to fix this.

Also, make sure you understand how mobile first works and what media queries are needed to apply it. You may find [this article](https://www.freecodecamp.org/news/taking-the-right-approach-to-responsive-web-design/) helpful 😉

Have a look at my comments below but again thumbs up for the effort 👌

index.html
askinekincihf: Thanks a lot for the review Antigoni. 

Added accessibility attribute .

```
<body>
    <header class="header">
      <div class="header-container">
        <a class="header-img"><img src="/img/logo.jpg" alt="logo"/></a>
```

style.css
askinekincihf: Yes, you are right, it was unnecessary, deleted the max-width.

`@media (min-width: 600px) `

style.css
askinekincihf: many thanks.

style.css
askinekincihf: I am supposed to think of the accessibility more.  Deleted  italic font style on body.

index.html
askinekincihf: typed class twice by mistake.

Amended the class name properly. 

`<div class="footer-list-policies">`

style.css
askinekincihf: I thought that auto is available for both margin and padding.

index.html
askinekincihf: amended pathway of logo source.

PR #96, opened 2020-12-17
London Class 7-Yasemin Bayraktutan-HTML/CSS-Week 3

ybayraktutan:
# Your Details

- Your Name: Yasemin Bayraktutan
- Your City: London
- Your Slack Name: Yasemin

# Homework Details

- Module: HTML/CSS
- Week:3


css/style.css
```diff
@@ -0,0 +1,405 @@
+body {
+    font-family: 'Josefin Sans', sans-serif;
```
LucyMac, line 2: Nice font 🙂

css/style.css
```diff
@@ -0,0 +1,405 @@
...
+
+.hamburger-menu {
+    align-self: flex-end;
+    margin-top: 10px;
+    margin-right: 20px;
+}
+
+.hamburger-menu img {
+    width:100%;
+    height: 20px;
```
LucyMac, line 27: You shouldn't need this height

css/style.css
```diff
@@ -0,0 +1,405 @@
...
+}
+
+.logo-icon {
+    width: 30%;
+    align-self: flex-start;
+}
+
+.logo-icon img {
+    width: 90%;
+    height: 90%;
```
LucyMac, line 45: Try to stick to only setting a width to your images, not a height.

css/style.css
```diff
@@ -0,0 +1,405 @@
...
+
+    .footer-text {
+        font-size: 1.1em;
+        text-align: center;
+    }      
+
+}
+
+/* Desktop design */
+@media (min-width: 900px) {
```
LucyMac, line 266: Good work using 2 breakpoints and following the mobile-first approach 👍

css/style.css
```diff
@@ -0,0 +1,405 @@
...
+}    
+
+/* Main Third Section */
+.container {
+    width: 100%;
+    display: flex;
+    flex-direction: row;
+    justify-content: center;
+    align-items: center;
+    flex-flow: wrap;
```
LucyMac, line 136: These alignment rules ensure that the images are always visible, but doesn't fully control how they are laid out. Sometimes, there is just one image by itself on the second line. 
To avoid this, make sure the images are small enough to fit on one row, or set `flex-direction: column` until they have enough room to be in a row.

LucyMac (commented):
Nice work Yasemin! Good looking page, clear HTML structure and nice responsive CSS. 

A few comments to improve your 4 images alignment in all screen sizes but overall really nice job 👍

PR #122, opened 2020-12-19
London class 7 - Omar Sharif - HTML/CSS - Week 3

omarsharif-gr:
# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

### DIFF
--- a/index.html
+++ b/index.html
@@ -6,11 +6,21 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Responsive Cake webpage</title>
-    <!-- Add a link to your css file here -->
+    <link rel="stylesheet" href="style.css">
 </head>
 
 <body>
+    <header>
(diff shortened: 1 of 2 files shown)


clairesmi:
Hi Omar, this is looking really good, great job 👍 
Your HTML is well structured, I would just say that maybe you didn't need the extra div within the footer element as you don't appear to be using it in your CSS so the styles could go directly on the footer. 

 I like where you were going with making the page responsive, with a burger menu for the mobile site but you shouldn't need to put any JavaScript in your HTML file to achieve this - you can create two separate nav menus in your HTML file and then use media queries with the `display: none` property to show and hide each nav based on the size of the screen. 
 
On line 32 in your CSS file you have overwritten your `display: flex` property on `topnav a` with `display: block` and this may cause some unexpected results so keep an eye out for things like this going forward. Also just remember to keep your formatting and indentation tidy. 

Keep up the great work and feel free to slack me if you have any questions!

PR #123, opened 2020-12-19
London class 7 - Soraiya Muhammad - HTML/CSS - Week 3

soraiya:
Name: Soraiya Muhammad
City: London
Slack Name: @soraiya

- Module: HTML/CSS
- Week: 3


index.html
```diff
@@ -5,12 +5,73 @@
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
-    <title>Responsive Cake webpage</title>
+    <title>Soraiya's Cake webpage</title>
     <!-- Add a link to your css file here -->
+    <link rel="stylesheet" href="style.css">
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Cardo&family=Karla:wght@300&family=Lora&family=Vollkorn&family=Vollkorn+SC&display=swap"
```
LucyMac, line 12: You'll loading quite a few font families here but I can only see 2 being used in your CSS. Bear in mind that loading all these fonts in will slow down your page load time, so it's best to only include what you need.

index.html
```diff
@@ -5,12 +5,73 @@
...
+        rel="stylesheet">
 </head>
 
 <body>
     <!-- Add your markup here -->
+    <section>
+    <header>
+        <div><img class="logo" src="/img/logo.jpeg"></div>
+        <div>
+            <H1>Soraiya's Bakery</H1>
```
LucyMac, line 22: HTML tags are always lowercase: `h1`

index.html
```diff
@@ -5,12 +5,73 @@
...
+        </nav>
+    <br>
+   
+        <section class="text-box">
+            <h2>A Piece of Cake for All Tastes</h2>
+            <p>Beautifully hand crafted custom cakes, catered to all dietary needs</p>
+        </div>
+        <div>
+            <img class="main-image" src="img/main-cake.jpg" alt="cake">
+            <div class="credit">(credit: Delphine Hourlay)</div>
```
LucyMac, line 49: Could you use a text tag like `p` or `span` here?

index.html
```diff
@@ -5,12 +5,73 @@
...
+        </section>
+     
+   
+    <div><h3>Browse our Collections</h3></div>
```
LucyMac, line 53: Do you need to wrap your heading inside a div here? Avoid unnecessary markup.

index.html
```diff
@@ -5,12 +5,73 @@
...
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Cardo&family=Karla:wght@300&family=Lora&family=Vollkorn&family=Vollkorn+SC&display=swap"
+        rel="stylesheet">
 </head>
 
 <body>
     <!-- Add your markup here -->
+    <section>
+    <header>
+        <div><img class="logo" src="/img/logo.jpeg"></div>
```
LucyMac, line 20: Your logo isn't showing. Add a dot before the first slash: `./img/logo.jpeg`
This means "the `img` folder is at the same level as this current file (index.html)"

style.css
```diff
@@ -1 +1,345 @@
 /*  Add your styling here */
+
+body {
+  background-color: #f0d7d2;
+  color: #4285f4;
+  font-family: "Karla", sans-serif;
+  display: flex;
+  justify-content: center;
+  align-items: center;
```
LucyMac, line 9: Try and limit rules on the `body` to very generic and common styles, such as the background-color and font-family, but not flexbox rules (as it's unlikely your entire page should follow the same flex layout rules always)

style.css
```diff
@@ -1 +1,345 @@
...
+
+header {
+  background-color: rgb(255, 255, 255);
+  display: flex;
+  align-items: center;
+  position: fixed;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 80px;
```
LucyMac, line 20: Nice

style.css
```diff
@@ -1 +1,345 @@
...
+  user-select: none;
+}
+
+.nav-menu input {
+  display: block;
+  width: 40px;
+  height: 30px;
+  position: absolute;
+  top: -8px;
+  right: 9px;
```
LucyMac, line 58: This is ok but if your aim is to vertically center something, remember to try using flexbox rules first, before resorting to position: absolute and fixed spacing

style.css
```diff
@@ -1 +1,345 @@
...
+  font-size: 25px;
+  padding-top: 40px;
+  padding-right: 0px;
+  padding-bottom: 40px;
+}
+
+.nav-menu input:checked ~ ul {
+  transform: translateX(100%);
+}
+
```
LucyMac, line 114: Nice mobile nav 👌

style.css
```diff
@@ -1 +1,345 @@
...
+  color: #4285f4;
+  text-transform: uppercase;
+}
+
+p {
+  margin-top: 10px;
+  color: rgb(65, 47, 4);
+  font-size: 18px;
+  font-style: oblique;
+  width: 80%;
```
LucyMac, line 146: Are you sure you want all your paragraphs to be that same width? If not, target the specific paragraphs with a class.

style.css
```diff
@@ -1 +1,345 @@
...
+.img-container {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  flex-wrap: wrap;
+}
+
+.collections-img {
+  width: 190px;
+  height: 207px;
```
LucyMac, line 171: Don't set a width AND a height on your `img` as this will distort them.  Setting the width should be enough, and let the height set itself naturally.

To better control the width of the image:
- wrap it inside a `div`
- give that `div` a width (and an optional height)
- and then just `width: 100%` on the `img` (no height)

You do need your images to be the same ratio / same shape for this to work.

style.css
```diff
@@ -1 +1,345 @@
...
+    font-style: normal;
+  }
+
+  h3 {
+    font-size: 28px;
+  }
+
+  .main-image {
+    width: 800px;
+    height: 900px;
```
LucyMac, line 245: Your image is 800px width, but this is set for screen width 600px and above. Which means when the screen is at 600px (all the way up to 800px) width, the image is wider than the screen. 
You could set the image to `width: 100%` instead so it adapts to the screen width.

LucyMac (commented):
Some good code and some good understanding of what we've learned - but do try to use semantic HTML, and also watch out how you size images so it doesn't distort them.

PR #136, opened 2021-01-27
West Midlands Class 2 - Kawthar - HTMLCSS - Week 3

pddys:
# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

### DIFF
--- a/index.html
+++ b/index.html
@@ -7,10 +7,119 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Responsive Cake webpage</title>
     <!-- Add a link to your css file here -->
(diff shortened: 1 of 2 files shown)


pddys (commented):
Good job. This is a complex layout to create and you've managed to develop it using some advanced CSS and some smart use of flexbox.

The things you could improve upon is trying to be dry with your HTML and CSS and let features inherent to CSS such as specificity and the cascade do the work for you. This means greater use of classes instead of IDs, and being a bit more selective about which HTML you add.

Generally though this is good work, and it's good to see that you've challenged yourself by looking at CSS beyond your curriculum such as CSS grid.

PR #152, opened 2021-06-10
Northwest 4-Omer Yamak-HTML/CSS-Week3

omerffaruk:

# Your Details

- Your Name: Omer Yamak
- Your City: Manchester
- Your Slack Name: Omer

# Homework Details

- Module: HTML/CSS
- Week: 3

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

-----
[View rendered readme.md](https://github.com/omerffaruk/HTML-CSS-Coursework-Week3/blob/north-west-4/omer-yamak/readme.md)


index.html
```diff
@@ -1,16 +1,106 @@
...
+        </div>
+      </section>
+    </main>
 
-</html>
\ No newline at end of file
+    <footer>
+      <div class="contact" id="contact">
+        <p>Get in touch with us to experience the most delicious cakes</p>
+        <p>Email:<strong> forexample@cakes.com</strong></p>
```
MCompton96, line 101: Sometimes a cleaner way to put spaces in with html is to use a non-breaking space `&nbsp;`, so in this case it would look like `<p>Email:&nbsp;<strong>forexample@cakes.com</strong></p>`

style.css
```diff
@@ -1 +1,218 @@
...
+ul {
+    list-style-type: none;
+}
+
+a {
+    text-decoration: none;
+    color: #000;
+}
+
+h1, h2, h3, h4 {
```
MCompton96, line 22: I don't think its a good idea to have the font weight of all the h tags the same, the idea of distingushing between h1-h6 is mainly through the difference in font-weight as well as size

style.css
```diff
@@ -1 +1,218 @@
...
+    background: url("/images/cake images/american-heritage-chocolate-vdx5hPQhXFk-unsplash.jpg");
+    background-size: cover;
+    background-repeat: no-repeat;
+    margin-bottom: 3rem;
+    border-radius: 15px;
+    box-shadow: 3px 0px 3px, -3px 0px 3px;
+}
+
+/* Cake-images section */
+.wrapper-images {
```
MCompton96, line 129: Really great utilisation of grid here

style.css
```diff
@@ -1 +1,218 @@
...
+    }
+    .navigation {
+        display: flex;
+        justify-content: flex-end;
+        margin-top: 2rem;
+    }
+    .navigation li {
+        margin-left: 1rem;
+    }
+    .navigation a {
```
MCompton96, line 185: Nice use of transition!

index.html
omerffaruk: Thanks Matthew. I didn't know about the "&nbsp" in html. Will use it from now on

style.css
omerffaruk: I see. The designed seemed better to me with headings set to the  same font-weight. Will avoid doing it.

PR #242, opened 2021-12-02
LONDON CLASS-8 RAHWA GHEBREMICHAEL HTML/CSS WEEEK-3

rahwafesseha:

# Your Details

- Your Name: Rahwa Ghebremichael
- Your City: London
- Your Slack Name: Rahwa Ghebremichael

# Homework Details

- Module: HTML/CSS
- Week: 3

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?


index.html
```diff
@@ -1,16 +1,103 @@
...
+                          <figcaption>Vegan chocolate cake</figcaption>
+                        </div>
+                      </div>
+                </main>
+              <!-- footer section -->
+
+                    <footer class="footer"></p><button>Place your order</button>
+                      <hr>
+                          <div class='footer-copywrites'>
+                              <p class='footer-copywrites'>&copy 2021 @Rahwit</p>
```
mahsa2, line 88: It's better to be consistant in using `'` or `"` in the whole code.

index.html
```diff
@@ -1,16 +1,103 @@
...
+
+              <!-- Nav section for wide screen -->
+
+                <nav class="nav"> 
+                  <ul class="navbar">
+                    <li class="home "><a href="#">Home</a></li>
+                    <li class="cakes "><a href="#">Cakes</a></li>
+                    <li class="order"><a href="#">Ordering</a></li>
+                    <li class="lessons "><a href="#">Lessons</a></li>
+                    <li class="about "><a href="#">About</a></li></ul>
```
mahsa2, line 38: It's not wrong but we only use spaces in classnames to separate multiple classes. E.g.
```
class="about"
```
vs.
```
class="about gap"
```

style.css
```diff
@@ -1 +1,175 @@
...
+
+.welcome {
+  display: grid;
+  grid-template-areas:
+    "welcome-text welcome-text welcome-text welcome-text welcome-text welcome-text"
+    ". . welcome-image welcome-image  . .";
+}
+
+.welcome-image {
+  grid-area: welcome-image;
```
mahsa2, line 69: Really great job on using the grid areas 🥇

style.css
```diff
@@ -1 +1,175 @@
...
+    font-size: 20px;
+    color: white;
+  }
+  .welcome {
+    grid-template-areas: "welcome-image welcome-image welcome-image welcome-text welcome-text welcome-text";
+    grid-gap: 10px;
+    padding: 10px;
+    text-align: end;
+  }
+  .logo-text {
```
mahsa2, line 172: I'm so happy to see the great use of styling in your code.
Just one more note: you need to add one new line between each css rule:
```
.welcome{
   ...
}

.logo-text {
  ...
}
```

index.html
```diff
@@ -1,16 +1,103 @@
...
+                          <figcaption>fancy cake</figcaption>
+                        </div>
+                        <div class="product-4">
+                          <figcaption>Vegan chocolate cake</figcaption>
+                        </div>
+                      </div>
+                </main>
+              <!-- footer section -->
+
+                    <footer class="footer"></p><button>Place your order</button>
```
mahsa2, line 85: `<footer>` and `<main>` are siblings of the same parents, therefore, they should have the same indentations.

index.html
```diff
@@ -1,16 +1,103 @@
...
 
-<body>
-    <!-- Add your markup here -->
-</body>
+            
+        
 
-</html>
\ No newline at end of file
+
```
mahsa2, line 103: We don't want to leave extra new lines or even extra spaces left at the end of our files.
You can look for some plugins for the editors to do it automatically for you whenever you save your file.

style.css
```diff
@@ -1 +1,175 @@
...
+  font-weight: bold;
+}
+.grid-container {
+  grid-auto-rows: auto;
+  grid-template-areas:
+    "header header"
+    "main main"
+    "footer footer";
+  padding: 2em;
+  grid-gap: 1em;
```
mahsa2, line 23: None of these `grid` properties work here. Do you know why?

Answer: Because you forgot to set `display: grid` to enable the grid properties.

index.html
```diff
@@ -1,16 +1,103 @@
...
+                alt="strawberry cap cake cake"
+              />
+            </div>
+            <div class="logo-text"><em>The best cake in town delivered to your door.</em>
+            </div>
+          </header>
+
+              <!-- Nav section for wide screen -->
+
+                <nav class="nav">
```
mahsa2, line 32: `<nav>` should be inside the `<header>`.

style.css
```diff
@@ -1 +1,175 @@
...
+  grid-area: logo-image;
+  position: absolute;
+  right: 0%;
+  bottom: 10px;
+  height: 45px;
+  width: 50px;
+}
+
+.logo-text {
+  justify-content: flex-end;
```
mahsa2, line 54: `justify-content` won't work here, can you guese why?

Answer: `display: flex` is missing.

style.css
```diff
@@ -1 +1,175 @@
...
+  background-position: center;
+}
+
+.product-3 {
+  background-image: url("./images/fancy.jpeg");
+  background-size: cover;
+  background-repeat: no-repeat;
+  background-position: center;
+}
+.product-4 {
```
mahsa2, line 101: Here we have 4 similar classes with almost identical css rules. We definetly want to avoid it. We want to reuse as much classes as possible for identical shaped items.

If you see that you need to give numbers to your classes like here (e.g. `product-1`, `product-2`, ...), it's a sign that you need to reuse css class.

I.e.

```
.product {
  background-size: cover;
  background-repeat: no-repeat;
  background-position: center;
}
```

But, in this case, it would have been better to use `` rather than a `background-image` because the images are not really a background but an image describing the products on the page. Read more about their differences: https://stackoverflow.com/a/1469139/1189105

Also, with `img` tag you won't need separate classes anymore for sure. Then you can use `object-fit` and `object-position` to place your image in the right location: 
https://www.w3schools.com/cssref/tryit.asp?filename=trycss3_object-fit2

index.html
mahsa2: Also, `</p>` needs to be removed here.

style.css
```diff
@@ -1 +1,175 @@
...
+  font-weight: bold;
+}
+
+button {
+  padding: 18px;
+  background: linear-gradient(orange, white);
+  color: black;
+  border-radius: 5px;
+  border-style: none;
+  justify-items: center;
```
mahsa2, line 126: `justify-items` will only work if the `display` is `flex` or `display`. But also depends on its value, it will work on either of the two.

mahsa2: If you want to put the button in the center, you need to define a parent div around the `button` and use `flex` to center the button:
```
wrapper: {
    display: flex;
    justify-content: center;
}
```

mahsa2 (commented):
Rahwa well done. The website looks great and I'm happy that you've improved the tips that we discussed in the past. It's always great to see your progress. 💯 🥇 

I left some comments here and I strongly recommend you to look at them.

PR #248, opened 2021-12-03
london8-Monika-Dangol-week3

Monika39-maker:

# Your Details

- Your Name: Monika
- Your City: London
- Your Slack Name: Monika Dangol

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?


style.css
```diff
@@ -1 +1,239 @@
...
+    
+
+}
+/* .......welcome/intro style finish...... ............ */
+
+
+/* ..........main/big image div style start..........  */
+.big-image {
+    height: 300px; 
+    width: 100vw;
```
beckyf0keefe: There is a bit of an issue with your page being wider than the viewport, the scroll bar horizontally is not something we want unless we designed it that way! The issue while not in desktop mode is that you set the width to be 100vw _buuut_ you already have padding from parent containers, this means that it will be 100vw + padding, which means you get the scroll bar, try to avoid setting things to 100vw unless you know they aren't a child of containers with styling.

style.css
```diff
@@ -1 +1,239 @@
...
+    .big-image {
+        grid-column: 1/2;
+        height: 300px;
+        width: 60vw;
+        
+    }
+
+    .intro-texts {
+        grid-column: 2/4;
+        font-weight: 100;
```
beckyf0keefe, line 218: Nice seeing this section move when the screen grows, good adaptivity

style.css
```diff
@@ -1 +1,239 @@
...
+    }
+    
+    nav {
+       grid-row: 2/3; 
+    }
+
+    .big-image {
+        grid-column: 1/2;
+        height: 300px;
+        width: 60vw;
```
beckyf0keefe, line 212: This is the same as in the desktop version, .big-image is 60vw and .intro-texts is also 40vw so overall you have your 100vw plud padding, hence the scroll bar.

index.html
```diff
@@ -7,10 +7,63 @@
...
     <title>Responsive Cake webpage</title>
     <!-- Add a link to your css file here -->
+    <link rel="stylesheet" href="style.css">
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.0.0-beta3/css/all.min.css" integrity="sha512-Fo3rlrZj/k7ujTnHg4CGR2D7kSs0v4LLanw2qksYuRlEzO+tcaEPQogQ0KaoGN26/zrn20ImR1DfuLWnOo7aBA==" crossorigin="anonymous" referrerpolicy="no-referrer" />
 </head>
 
 <body>
     <!-- Add your markup here -->
+    <div class="container">
+        <header>
```
beckyf0keefe, line 17: Overall, nice seeing some semantic HTML being used, where do you think we could use a couple more, where do you thing main would go? Would a section or article be appropriate for some parts of your page?

style.css
```diff
@@ -1 +1,239 @@
...
+
+
+/* ...........small images div style start.......  */
+.small-images {
+    grid-column: 1/-1;
+    background: pink;
+    display: flex;
+    flex-flow: row wrap;
+    justify-content: space-evenly;
+    align-items: space-between;
```
beckyf0keefe, line 140: nice use of flex!

style.css
```diff
@@ -1 +1,239 @@
...
+
+
+/* ..........main/big image div style start..........  */
+.big-image {
+    height: 300px; 
+    width: 100vw;
+    background-image: url("../images/cake2.jpg");
+    background-position: center;
+    background-size: cover;
+    background-repeat: no-repeat;
```
beckyf0keefe, line 124: Do you think that this should be a background? or an image tag? What is it the background to? If it was an image tag we could add an alt tag for users who can't see the images, which is really important

index.html
```diff
@@ -7,10 +7,63 @@
...
+        </div>
+        
+        <div class="big-image">
+            <!-- img inside this div  -->
+        </div>
+        <div class="small-images">
+            <img src="../images/cake1.jpg" alt="" class="small-img">
+            <img src="../images/cake3.jpg" alt="" class="small-img">
+            <img src="../images/cake4.jpg" alt="" class="small-img">
+            <img src="../images/cake5.jpg" alt="" class="small-img">
```
beckyf0keefe, line 53: What are the alt tags for these images? make sure you add them, otherwise it won't pass accessibility requirements

beckyf0keefe (commented):
Overall looking good, good use of flex and grid in together making a properly responsive page. Be careful about making sure the page width doesn't exceed the view-port but once I changed that, it was looking really good 👍

PR #256, opened 2021-12-03
London Class 8 - Shadab Safaee - HTML CSS Coursework - Week3

migmow:
- Your Name: Shadab Safaee 
- Your City: London
- Your Slack Name: Shadi

# Homework Details

- Module: HTML-CSS-Coursework
- Week:  3

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?


main.css
```diff
@@ -0,0 +1,189 @@
...
+        grid-row-gap: 50px;
+    }
+
+    .nav {
+        display: block;
+        grid-column: 2/12;
+        grid-row: 2/3;
+        padding: 1em;
+        text-align: right;
+        font-size: 2.2vw;
```
beckyf0keefe, line 124: Just be careful with this, the height of the buttons is being dictated by font size, where the size of the nav container is being dictated by the height of the row, this means that when the screen gets wider, the buttons grow past the height of the container. The easy way to get around this is by using rem instead of vw for fontsize.
image

index.html
```diff
@@ -6,11 +6,109 @@
...
+            <div class="burger_item"></div>
+            <div class="burger_item"></div>
+        </div>
+
+    </header>
+
+    <main class="main">
+
+        <nav class="nav">
+            <ul class="tab-content">
```
beckyf0keefe, line 39: nice use of semantic HTML

main.css
```diff
@@ -0,0 +1,189 @@
...
+
+    .nav ul li a {
+        color: black;
+        text-decoration: none;
+    }
+
+    .hero {
+        grid-row: 3/6;
+        display: flex;
+    }
```
beckyf0keefe, line 149: I wasn't sure what you were doing with the flex and grid at first for the hero but I like this way of doing it!

main.css
```diff
@@ -0,0 +1,189 @@
...
+
+.glutenfree_product {
+    grid-column: 3/5;
+    grid-row: 2/3;
+    background: black;
+    background-image: url(image/Pexel/choc\ cake.jpg);
+    background-size: cover;
+    background-position: center;
+    cursor: pointer;
+}
```
beckyf0keefe, line 94: Should we be using the background property here? What are these images a background of? If we don't use the img tag, we can't add alt tags and when you don't add tags to images (that aren't decorative) then you won't meet accessibility standards.

index.html
```diff
@@ -6,11 +6,109 @@
...
+            <section class="hero_text">
+                <h1>Welcome</h1>
+                <p>
+                    Enjoy a wide range of savoury cakes served right at your doorstep. With our easy to use online shop,
+                    you will be spoilt with many choices of freshly baked cakes!
+                </p>
+            </section>
+
+            <section class="hero_img">
+                <img src="image/best_cupcakes.jpg" alt="range of cakes">
```
beckyf0keefe, line 59: Good to see you using alt-text!

index.html
```diff
@@ -6,11 +6,109 @@
...
+    <link rel="stylesheet" href="header.css">
+    <link rel="stylesheet" href="main.css">
+    <link rel="stylesheet" href="footer.css">
+
 </head>
 
 <body>
-    <!-- Add your markup here -->
+
+    <header class="header">
```
beckyf0keefe, line 18: Careful with not having the text and the burger menu overlap on smaller screen.

beckyf0keefe (commented):
Overall looking good, good job on the hero, everything was nicely responsive, just be careful with making sure your text isn't overlapping/forcing buttons out of their container, how might we fix the issue with the buttons in the nav bar?

PR #264, opened 2021-12-03
LONDON CLASS_8 - DEAGO BROWNE - HTML/CSS - WEEK_3

DeagoBrowne:

# Your Details

- Your Name: **Deago Browne**
- Your City: **London**
- Your Slack Name: **Deago Browne - Trainee**

# Homework Details

- Module: **html/css**
- Week: **3**

# Notes

- What did you find easy? **funnily enough, the media queries were the easiest things to implement.**

- What did you find hard? **Getting my head around CSS Grids, and how it interacts with Flexbox**

- What do you still not understand? **How to get an image to fill the entire container height-wise, when contained within a flex box that also contains text - always seems to be a little section at the bottom that is visible if I change the background colour** 

- Any other notes? **I quite enjoyed this assignment - but it is a shame that I caught the flu, as I lost a lot of time, and handed in my assignment 3hrs after the deadline.**


style.css
```diff
@@ -1 +1,193 @@
...
+.footer-text {
+    font-family: 'Estonia', cursive;
+    font-size: 30px;
+    text-align: center;
+}
+
+
+@media screen and (min-width: 412px) {
+    header {
+    background-size: cover;
```
mahsa2, line 99: indentation is missing.

style.css
```diff
@@ -1 +1,193 @@
...
+    
+    .footer-text {
+        height:fit-content;
+        background-color: rgb(151, 207, 204);
+    }
+}
+
+@media screen and (min-width: 1503px) {
+
+    header {
```
mahsa2, line 185: We don't want to leave empty lines at the begining and ending of the class (ie after `{` and before `}`).

index.html
```diff
@@ -6,11 +6,49 @@
...
-    <!-- Add your markup here -->
+<body class="container">
+    <div></div>
+    <header class="header">
+            <div class="banner-text">
+                <h1>B.G. CAKES</h1>
+            </div>
+    </header>
+    <main class="welcome">
+        <div class="main">
```
mahsa2, line 23: To avoid confusion, we don't want to give a class name `main` to another div other than `main`. In face, you don't need to define any class for `<main>` at all and you can define a css rule directly for it. Like:

```
main {
   ...
}
```

style.css
```diff
@@ -1 +1,193 @@
...
+
+p {
+    font-family: 'Estonia', cursive;
+    font-size: 40px;
+    font-weight: 1000;
+}
+
+.main-img {
+    width: 100%;
+    height: 200px;
```
mahsa2, line 65: If you define `height` directly here, when the website resolution changes, the image gets distorted, we simply don't set anything and the default `height` would be automatically calculated here.

style.css
```diff
@@ -1 +1,193 @@
...
+@media screen and (min-width: 1025px) {
+    header {
+        height: 300px;
+        background-size: 1500px 500px;
+        margin-bottom: 20px;
+    }
+
+    .banner-text {
+        top: 125px;
+        left: 35%;
```
mahsa2, line 153: This `left` made your desktop version to get a horizontal scrolling which we want to avoid.

style.css
```diff
@@ -1 +1,193 @@
...
+
+   p {
+       font-size: 50px;
+   }
+
+
+    .img-container {
+        grid-template-columns: 1fr 1fr 1fr 1fr;
+
+        margin-bottom: -300px;
```
mahsa2, line 173: We don't need a new line between css rules. (line 172)

index.html
```diff
@@ -6,11 +6,49 @@
...
+    <link rel="stylesheet" href="style.css">
+    <style>
+        @import url('https://fonts.googleapis.com/css2?family=Comforter&family=Estonia&family=Praise&display=swap');
+        </style>
 </head>
 
-<body>
-    <!-- Add your markup here -->
+<body class="container">
+    <div></div>
```
mahsa2, line 16: Empty html tags should be removed if they don't have any css rules on them.

style.css
```diff
@@ -1 +1,193 @@
...
+}
+
+
+.gallery {
+    grid-area: gallery;
+}
+
+.gallery-img {
+    width: 100%;
+    height: 150px;
```
mahsa2, line 75: If you define an explicit `height` on your image, on screen resize, the image gets distorted.
One solutions here is this, because the grid will define the max size of each image. Otherwise, you need to define a wrapper around the image and give it explicit width and height.
```
.gallery-img {
   width: 100%;
   height: 100%;
   object-fit: cover;
}
```

style.css
mahsa2: You could use `justify-content: right` and then use margin-top to adjust it in the spot you want.

style.css
```diff
@@ -1 +1,193 @@
 /*  Add your styling here */
+
+body {
+    background-color: rgb(234, 239, 241);
```
mahsa2, line 4: By defining `margin: 0` on `body` you make sure that there is no unwanted spacing around your site, unless you want it to be. In that case you need to use `padding` or `margin` on the elements.

index.html
```diff
@@ -6,11 +6,49 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Responsive Cake webpage</title>
-    <!-- Add a link to your css file here -->
+    <link rel="stylesheet" href="style.css">
```
mahsa2, line 9: All the html tags `link`, `meta`, here needs to be self-closed using `/`:
```
<meta ... />
<link ... />
```

index.html
```diff
@@ -6,11 +6,49 @@
...
+                <img src="./img/syed-maaz-QLSh4y2v93c-unsplash.jpg" alt="Designer Cake" class="main-img"/>      
+            </div>
+        </div>
+    </main>
+    <section class="gallery">
+        <div class="img-container">
+            <img src="./img/mitchell-luo-LASbBiXPO5I-unsplash.jpg" alt="A delicious chocolate fudge cake with strawberries and chocolate drizzle" class="gallery-img">
+            <img src="./img/rakesh-sitnoor-vsBF9EqxRno-unsplash.jpg" alt="A gorgeous cake on a gold platter" class="gallery-img">
+            <img src="./img/slashio-photography-2k9CSieuABw-unsplash.jpg" alt="A beautiful cake made to look like a rose" class="gallery-img">
+            <img src="./img/tanya-pro-XTI0pEqQ8DI-unsplash.jpg" alt="A wonderful purple and white cake" class="gallery-img">
```
mahsa2, line 40: All the `img` tags should be self-closed with `/`.
```
<img ... />
```

index.html
```diff
@@ -6,11 +6,49 @@
...
+                <p>Welcome to the most BADMAN cake delivery service that you have EVER
+                    encountered. <br> <br>Our cakes are SO GANGSTER, that you won't even want to eat them!
+                </p>
+            </div>
+            <div>    
+                <img src="./img/syed-maaz-QLSh4y2v93c-unsplash.jpg" alt="Designer Cake" class="main-img"/>      
+            </div>
+        </div>
+    </main>
+    <section class="gallery">
```
mahsa2, line 35: `<main>` mostly should be around the main part of the page except footer and header. The `section` should be in the `<main>`.

mahsa2 (commented):
Nice work Deago. The website looks great. Happy to see your progress. Also well done on the `grid` use. 💯 🥇 

Don't forget to read through the comments, as it's an important part of learning.

PR #274, opened 2021-12-04
London Class 8 - Delroy Gayle - HTML/CSS - Week 3

DelroyGayle:
Done!

- What did you find easy?
Cannot say that I found any of it easy.

- What did you find hard?
Wrapping my head around the concept of CSS Grids. I kept having to restart & rethink whilst I was coding. Because it became apparent that I had misunderstood the concept.

- GITHUB!!! I even struggled to submit my work. I cannot even tell what I did now correctly in order to submit? Beforehand Github kept saying there was nothing to compare.

- Any other notes?
Are there any good online tutorials regarding CSS Grids?


index.html
```diff
@@ -6,11 +6,211 @@
...
     <title>Responsive Cake webpage</title>
-    <!-- Add a link to your css file here -->
+    <link rel="stylesheet" href="style.css" />
 </head>
 
 <body>
-    <!-- Add your markup here -->
+    <section class="homepage-container">
+        <section class="hero-box">
+                <section class="hero-flexbox">
```
LucyMac: What happened to using `<header>`, `<main>` and `<footer>`? Please edit your HTML to use a semantic structure. 
Having `section`s inside `main` is good to split the main part of the page into logical blocks, but these must all live inside the `main` tag

index.html
```diff
@@ -6,11 +6,211 @@
...
+                                    <input type="checkbox" id="hamburger"/>
+
+                                <!-- (C) MENU ITEMS -->
+                                <div class="hamitems">
+                                    <a href="#">HOME</a>
+                                    <a href="#">CAKES</a>
+                                    <a href="#">ORDERING</a>
+                                    <a href="#">LESSONS</a>
+                                    <a href="#">ABOUT US</a>
+                                </div>
```
LucyMac: Nice mobile nav 👍 
- Please use `<ul>` and `<li>` tags for lists of items like this. You've done this really well for the desktop version of the nav - this should be no different. 
- Can you use CSS to make this mobile nav lay _on top_ of the homepage content, rather than pushing everything down when it's open?

index.html
```diff
@@ -6,11 +6,211 @@
...
+                            </div>
+                        </section>
+
+                        <!-- <section class="desktop-heading">
+                            <h1 class="mainheading2">ASTERISK ET OBELIX<br/>CAKES!</h1>
+                           <br/>
+                           <div class="thequote2">
+                               <p><em>We Just Do Great Cakes!<br/><br/>The best cakes in town delivered to your door</em></p>
+                           </div>
+                        </section> -->
```
LucyMac: There's a lot of commented out code in here. Avoid this. 
Comments to explain something or make a mental note is fine - but not commenting out actual code that you've decided not to use.

index.html
```diff
@@ -6,11 +6,211 @@
...
+                <p class="desktop-prompt">
+                    We make all types of same-day cakes
+                    </br> from Cupcakes, Celebration Cakes,
+                    </br> Birthday Cakes, Chocolate cakes, Novelty cakes, Bespoke cakes
+                    </br> and even same-day Wedding Cakes.
+                    <br/><br/>If you can dream of it we can make it.<br/><br/>
+                    
+                    Hover over the image to the left and the images below for more information
+                </p>
+
```
LucyMac: Why are you repeating all this content 3 times here? You only need it once. If you need to adapt the styling for each device, use media queries to change the style.

Secondly, do not use `<br>` to force the content onto the next line. Remove all the `br`s and let the content flow naturally and adapt to the space it has on each screen size.

index.html
```diff
@@ -6,11 +6,211 @@
...
+                        </figure>
+
+                        <figure class="hoverwrap">
+                            <img src="img/chocolate-cakeHW3.png" class="display-a-cake" alt="A Chocolate Cake">
+                            <!-- THE HOVER TEXT: -->
+                            <div class="hovercap">We are a little health-conscious and like to make sure our cakes are irresistibly delicious
+                                                   without using too much fat and sugar "just as much as needed to make something delicious". 
+                            </div>
+                        </figure>   
+                    </section>
```
LucyMac: You do not need to group your images 2 by 2 here. Put them all in the same container, and then use CSS Grid to align the items 2 by 2 on mobile, and then all in a row on larger screens. It's be much cleaner and neater HTML, with just one parent and 4 child items, and your CSS should be more straight forward as well.

LucyMac (commented):
A good start @DelroyGayle, and really not as bad as you let on :) 
BUT you didn't use CSS Grid! Please re-work your 4-image section to use grid - see my comments below. 
Nice work using some JS in there 👍

style.css
```diff
@@ -1 +1,727 @@
...
+    "cake      cake         cake        cake"
+    "flex       flex        flex        flex"  
+    "footer     footer      footer      footer";
+} 
+
+
+
+
+/* TABLET: */
+@media screen and (min-width: 481px) {
```
LucyMac: I'd start tablet at a larger size - somewhere between 600px and 800px

style.css
```diff
@@ -1 +1,727 @@
...
+      font-size:14px;
+      color: mistyrose;
+      text-align:center;
+  }
+
+
+/* THE GRIDS AND MEDIA QUERIES START HERE */
+
+.homepage-container {
+    width: 320px;
```
LucyMac: I wouldn't set a width on your page like this. Perhaps a `min-width` in this case?
It needs to fill the space if the phone is 450px wide for example, which a min-width will allow it to do.

style.css
```diff
@@ -1 +1,727 @@
...
+
+   /*  
+    We only define rules which are DIFFERENT from the mobile version defined above. 
+  */   
+
+          .homepage-container {
+            margin:3em;
+            width:100%; /* Use the full screen */
+            display: grid;
+            grid-template-columns: repeat(4, 1fr);
```
LucyMac: You don't need to repeat the rules that haven't changed from the mobile version. Only over-write those which have changed.

style.css
```diff
@@ -1 +1,727 @@
...
+    THE ACTUAL SIZE OF THE IMAGE IS 400 x 360 PIXELS, 37.44 x 27.44 rem 
+    */
+
+          .AsterixetObelix {
+              width:30em; /* Increase taking advantage of more space */
+              margin-top:2em;
+          }
+
+          .thequote {
+            font-size:300%;
```
LucyMac: Prefer `rem` units for font-size.

style.css
```diff
@@ -1 +1,727 @@
...
+  
+}
+
+/* Otherwise: */
+@media screen and (min-width: 481px) {
+
+      .mainheading{
+          font-size:400%;
+          text-align:right;
+          margin-left:12em;
```
LucyMac: This huge margin is breaking your layout - remove it.
You shouldn't need it - something like setting the header to flex and using `space-between` should naturally keep the logo to the left and the heading to the right.

PR #365, opened 2022-08-09
NW_5-Manchester-Ahmed_Mahmoud-HTML-CSS-Week3

Danti8686:

# Your Details

- Your Name: Ahmed Mahomud
- Your City: Manchester
- Your Slack Name: Danti

# Homework Details

- Module: HTML-CSS
- Week: 3

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?


style.css
```diff
@@ -1 +1,285 @@
...
+
+    }
+
+    .box {
+        padding: 21px;
+    }
+
+    .container {
+        grid-template-columns: 1fr 1fr 1fr 1fr 1fr 1fr;
+        grid-template-rows: 1fr 1fr 1fr 1fr 1fr 1fr;
```
yogaraptor, line 130: Having the rows all the same height means that on larger screens the header is very tall, even though it doesn't need to be. Here's a screenshot of it on a full HD screen (1920x1080px, which is a common screen size for laptops)

image

I'd suggest rather than trying to define one grid for the whole page, just let the sections of the page (header, welcome, article, images, etc) stack on top of each other naturally, and then define grids within those elements when you need to. This means you don't need to think about how each row affects the whole page, and can treat them as isolated layouts.

style.css
```diff
@@ -1 +1,285 @@
...
+    border: 3px solid black;
+}
+
+.container {
+    display: grid;
+    grid-template-columns: repeat(2, 1fr);
+    grid-template-rows: 0.3fr 0.3fr 0.5fr 0.5fr 0.5fr 0.5fr 0.5fr;
+    grid-gap: 2px;
+
+    grid-template-areas:
```
yogaraptor, line 11: Nice experimentation with named areas as opposed to using `grid-column-start` etc (which you do at the larger breakpoint). If you were building this "for real" then I'd suggest sticking to _either_ named areas or `grid-column-start` and related properties, so it's easy to reason about how the layout changes between the two breakpoints.

style.css
```diff
@@ -1 +1,285 @@
...
+    }
+
+    .box {
+        padding: 21px;
+    }
+
+    .container {
+        grid-template-columns: 1fr 1fr 1fr 1fr 1fr 1fr;
+        grid-template-rows: 1fr 1fr 1fr 1fr 1fr 1fr;
+        grid-area:
```
yogaraptor, line 131: `grid-area` is for assigning an element to a particular area within a grid, were you thinking of `grid-template-areas` here? Like you did [here](https://github.com/CodeYourFuture/HTML-CSS-Coursework-Week3/pull/365/files#diff-b78be019f1dc6d57753ea900c3805b114cd53ab7c0db836cc081836df1b99b7aR11)?

This actually isn't causing any problems, because the browser will discard the property as it doesn't have a valid value, but you can remove this for tidiness.

Hint: you can see when a CSS property isn't valid in the browser's dev tools:

image

yogaraptor (commented):
Hey :wave: left a couple comments, but nice use of CSS grid to totally change the layout of the page between breakpoints!

A few general points:

- You don't need to worry about making your class names, grid areas short (e.g. `mimg`). When a web server sends your page down to a user's browser it will compress it, so feel free to make these names longer and more descriptive (e.g. `middle_image` instead of `mimg`). This will help others (and you!) reading your code.
- For your next site, try adding some hover effects to your interactive elements (e.g. nav links). Little touches like this can help the user stay engaged. I see you did the [UI design tutorial](https://scrimba.com/learn/design) as well - next time try and apply a few of your favourite tricks from there too. 

But again - really good to see you experimenting with grids and trying the different ways of using them. Keep exploring :muscle:

index.html
```diff
@@ -6,11 +6,91 @@
...
+    <div class="container">
+
+        <div class="icon_img box">
+            <img src="https://www.bing.com/th?id=OIP.5wHPVNq-5T5Adqavx62ouwAAAA&w=150&h=150&c=8&rs=1&qlt=90&o=6&pid=3.1&rm=2"
+                alt="icon" width="100px">
+        </div>
+
+        <div class="bor">
+            <ul class="bor_li box">
+                <li class="">Home</li>
```
yogaraptor, line 22: I know this site is mainly for demoing grid, but remember these would be links "in real life"

index.html
```diff
@@ -6,11 +6,91 @@
...
+            </ul>
+        </div>
+
+        <div class="p_nav box">
+            <h3>The best cakes in town<br>delivered to your door</h3>
+        </div>
+
+        
+
+        <div class="wel_tex box">Welcom
```
yogaraptor, line 35: Make sure you've [got the Prettier extenstion installed](https://docs.codeyourfuture.io/Leaders/joining-the-course/getting-setup#5.2-extensions) and ["format on save" enabled in your editor](https://docs.codeyourfuture.io/Leaders/joining-the-course/getting-setup#5.2-extensions), to tidy up your code - e.g. here it would move the `Welcom` text onto the next line and indent it correctly.

index.html
```diff
@@ -6,11 +6,91 @@
...
+
+
+        <div class="main_img box">
+            <img src="https://th.bing.com/th/id/OIP.hT1QYACXEUqs6uTy8DyC6AHaE7?w=247&h=180&c=7&r=0&o=5&pid=1.7"
+                alt="main cakes" height="100%" width="100%" />
+        </div>
+
+        <div class="fimg_pic box">
+            <img src="https://th.bing.com/th/id/OIP.ylWiJppALlJ_lrxBgSJMWwHaJ4?w=186&h=248&c=7&r=0&o=5&pid=1.7"
+                alt="img1-1" height="100%" width="100%">
```
yogaraptor, line 62: Remember that alt text should provide a textual representation of the meaning of the image. So in this case it would describe the cake in the photo.

PR #383, opened 2022-08-12
NW5-Manchester-Zaw Khing-HTML/CSS-Week 3

zkhing:

# Your Details

- Your Name:Zaw Khing
- Your City:Manchester
- Your Slack Name:Zaw Khing

# Homework Details

- Module:HTML/CSS
- Week:3

# Notes

- What did you find easy?
- Adding new font

- What did you find hard?
- Css grid

- Css positioning and layout

- Any other notes?

### DIFF
--- a/index.html
+++ b/index.html
@@ -45,16 +45,12 @@
 
         <p>The best cakes in town delivered to your door</p> 
 
-        
-
       </div>
     </header>
 
      
 
-<main>
-
-    <!-- <div class="container"> -->
+   <main>
 
        <div class="hero-info-img">
           <h1>Welcome</h1>
@@ -88,24 +84,23 @@
 
         </div>
         
- </main>
+  </main>
 
-    <footer>
-        
-
-         <div class="nav-bar">
-             <ul>
+  
+  <footer>
+        <div class="nav-bar">
+              <ul>
                   <li><a href="#">HOME</a></li>
                   <li><a href="#">CAKES</a></li>
                   <li><a href="#">ORDERING</a></li>
                   <li><a href="#">LESSONS</a></li>
                   <li><a href="#">ABOUT</a></li>
-              </ul>
+               </ul>
             </div>
          <p>Authour: Zaw Khing</p>
          <a href="mailto:zawkhing@hotmail.com">zawkhing@hotmail.com</a>
 
-    </footer>
+  </footer>
 </body>
 
 </html>
--- a/index.html
+++ b/index.html
@@ -62,25 +62,19 @@
         <!-- <div class="hero-section"> -->
         <div class="allImages">
            
-            <div class="hero-image1 fourCakes">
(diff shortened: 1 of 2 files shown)


style.css
```diff
@@ -1 +1,180 @@
...
+
+.heroImages {
+  width: 100%;
+  height: 200px;
+}
+
+.hero-info-img h1 {
+  margin-left: 20px;
+  font-size: 50px;
+  color: #bde0fe;
```
Dominic-Taylor-Dev, line 132: One way you can check out the contrast is by using Chrome Dev Tools and hover over elements (icon to click is top right in this screenshot, or command + shift + C on Mac):

image

You can see that in this case the contrast between text and background is a little bit low - it would improve accessibility to have more contrast

style.css
```diff
@@ -1 +1,180 @@
...
+  justify-content: space-between;
+  background-color: #cdb4db;
+  width: auto;
+  height: 300px;
+  /* border: 1px solid black; */
+}
+
+.nav-img {
+  width: 100%;
+  height: auto;
```
Dominic-Taylor-Dev, line 45: I think this combination is what is making the image scale in a slightly strange-looking way - it's effectively locking the aspect ratio of the original image. Here's a screenshot of what it looks like on a certain size:

image

There's no easy or obvious answer for me here - it depends how you want the image to behave, but you could try something like:

```
.nav-img {
  width: 100%;
  height: 100%;
  object-fit: cover;
  object-position: 100% 0;
}
```

`object-fit: cover` would preserve the original aspect ratio, but the image still occupies all the available space.

 `object-position: 100% 0` makes it display the right hand side of your image first - so you see the right hand side as long as possible when scaling down (which I think was the most interesting part of your chosen image)

index.html
```diff
@@ -7,10 +7,94 @@
...
+        <a href="#" class="toggle-button">
+            <span class="bar"></span>
+            <span class="bar"></span>
+            <span class="bar"></span>
+            <span class="bar"></span>
+            <span class="bar"></span>
+        </a>
+
+        <div class="nav-bar"> 
+          <nav>
```
Dominic-Taylor-Dev, line 35: Some nice uses of semantic HTML here - good job :)

index.html
```diff
@@ -7,10 +7,94 @@
...
+
+       <div class="hero-info-img">
+          <h1>Welcome</h1>
+          <p>We are family run bakery.We can make any kind of cake for your birthday party or wedding or specail occasions. You can order it by phone or thorugh our online form.</p>
+          <img class="heroImg" src="https://media.istockphoto.com/photos/strawberry-and-cream-sponge-cake-on-white-wooden-table-picture-id1294578109?b=1&k=20&m=1294578109&s=170667a&w=0&h=cK7uKAhs3r08JVUA22omHA-ToHRr5duIRagkoJwBJHc=">
+        </div>
+
+        
+        <!-- <div class="hero-section"> -->
+        <div class="allImages">
```
Dominic-Taylor-Dev, line 63: Try to stick to a single convention on class naming (in CSS, usually it's hyphen separation - the 'camel case' you've used here is more of a JavaScript convention).

Here's one common way of structuring your names: http://getbem.com/naming/

Dominic-Taylor-Dev (changes requested):
Overall I think you did a great job here. Let's have a look at using CSS grid when we speak later :)

index.html
zkhing: Thank you Dom. I will make changes accordingly.

PR #385, opened 2022-08-12
KamNik0044:north-west-class-5/kam-nik

KamNik0044:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,16 +1,62 @@
 <!DOCTYPE html>
 <html lang="en">
-
 <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Responsive Cake webpage</title>
-    <!-- Add a link to your css file here -->
+    <link rel="preconnect" href="https://fonts.googleapis.com">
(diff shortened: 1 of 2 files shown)


SteveLeicester:
Hello Kam,

This is a nice piece of work.

You have used semantic HTML and divided your page up into smaller pieces.
The navigation bar is nicely done. Did you consider having that inside your header.  Often if you are building lots of pages they have the same header and navigation.  So you can often reuse the same header HTML in all your pages.

SteveLeicester:
Hello Kam,

This is a nice piece of work. You have used semantic HTML to break up you page into smaller pieces.  The navigation bar is nice. Did you consider putting this inside the header?  One reason is that if you are building a website with many pages the header is often the same and also the navigation is often the same.  If the navigation is inside the header than it is just one thing that can be reused on all your pages.  It's not a strict rule though.

You have a main tag also which holds all the important information on your page.  You should then consider breaking your main content into sections rather than having a section separated from main.

The footer is good. You need to put your social medial links there.

Finally did you try mobile first approach? So for a small device all content can be in a single column.  Perhaps each cake image in it's own section?

Keep it up Kam
Steve

PR #405, opened 2022-11-07
London 9 Turing - susan - HTML/ CSS - Week 3

susanssky:
https://cyf-london9-turing-htmlcss3-hkuhku609.netlify.app/


index.html
```diff
@@ -1,16 +1,78 @@
...
+        <img src="./assets/3.jpg" alt="3" />
+        <img src="./assets/4.jpg" alt="4" />
+      </section>
+    </main>
+    <footer>
+      <p>Lorem ipsum dolor sit.</p>
+      <p>Lorem ipsum dolor sit.</p>
+      <p>Lorem ipsum dolor sit.</p>
+      <p class="susan">Coding by Susan</p>
+    </footer>
```
samhoooo, line 76: On desktop, the footer is not at the bottom. Any way to fix this?

image

samhoooo (commented):
Thanks your work! I have only one comment here.

index.html
susanssky: @samhoooo My linux laptop shows the footer at the bottom....
Screenshot from 2022-11-15 21-59-33

samhoooo: Is it related to the screen height? What if you test with dev tool?

susanssky: I have tested with "toggle device toolbar", all devices are no problem...

samhoooo: If you test with 1920 x 1080, can you see the problem?
image

susanssky:
@samhoooo Thank you for pointing out the problem. I have fixed this problem. Now it is no problem even if it is 2560×1440

PR #408, opened 2022-11-08
London Class 9 - Lovelace - Mo Mahboobian - HTML/CSS - Week 3

momahboobian:

## [Live Demo](https://cyf-momahboobian-cakes.netlify.app)
# Your Details

- Your Name: Mo Mahboobian
- Your City: London
- Your Slack Name: momahboobian

# Homework Details

- Module: HTML/CSS 
- Week: 3

# Notes

- What did you find easy?
Designing mobile first
- What did you find hard?
Fitting images in a css grid box
Grid nesting (specially a flex inside grid with unordered list elements)
- Any other notes?
It was so challenging to add hamburger menu


ShayanMahnam:
wow, great job!!! well done mate!

momahboobian:
> wow, great job!!! well done mate! 

Thank you 🤩

css/style.css
```diff
@@ -0,0 +1,276 @@
...
+  margin-left: 10px;
+}
+.logo-img img {
+  display: grid;
+  width: 100px;
+  opacity: 50%;
+}
+
+.header-content {
+  grid-column: 2 / -1;
```
jonnywyatt, line 34: This does align the text correctly 👏 
An alternative way to get this text vertically aligned to the bottom of the `.header` container, is to set display: flex on the `.header`, and then use `align-self: flex-end` on .header-content

css/style.css
```diff
@@ -0,0 +1,276 @@
...
+  font-size: 20px;
+  font-weight: 300;
+  font-style: italic;
+  color: #979797;
+  direction: rtl;
```
jonnywyatt: The direction property is for styling languages that read right-to-left, like arabic.
To right-align english, use `text-align: right`
([more info on direction/ rtl](https://developer.mozilla.org/en-US/docs/Web/CSS/direction))

css/style.css
```diff
@@ -0,0 +1,276 @@
...
+  font-family: "Roboto", sans-serif;
+  font-weight: 300;
+}
+
+/* ----- Header ----- */
+
+.header {
+  display: grid;
+  grid-template-columns: 40% 60%;
+  background-color: #eee;
```
jonnywyatt, line 19: Putting hex colour values inline like this works fine 👍 
Even better is to use CSS variables to store the colours, will make them readable as you use them in your CSS. You can declare them once in a `:root` selector then re-use them ([example](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties#using_the_root_pseudo-class))

index.html
```diff
@@ -1,16 +1,112 @@
...
+  <body>
     <!-- Add your markup here -->
-</body>
+    <header>
+      <div class="header">
+        <div class="logo-img">
+          <img src="/img/logo-s.png" alt="Cake logo" />
+        </div>
+        <div class="hamburger-container">
+          <div class="hamburger">
```
jonnywyatt, line 26: Because this will be a clickable element that will use Javascript, the best tag to use is `button`. That's better for accessibility - eg it will let the user tab onto it using the keyboard. ([reference](https://developer.mozilla.org/en-US/docs/Learn/Accessibility/HTML#html_and_accessibility))
You'll need to remove the default padding, margin and border that comes with `button` though.

index.html
```diff
@@ -1,16 +1,112 @@
...
+          </p>
+        </div>
+      </section>
+      <figure class="card-img">
+        <img class="main-img" src="/img/slide1.jpg" alt="" />
+      </figure>
+
+      <section class="products">
+        <div class="product-item">
+          <img class="product-img" src="/img/prod1.jpg" alt="" />
```
jonnywyatt: Don't forget to set `alt` text for these images, your screen reader users will thank you!

index.html
```diff
@@ -1,16 +1,112 @@
...
+          <a href="mailto:yourmail@gmail.com"> yourmail@gmail.com </a>
+          <p>Telephone: (1)222 333 44</p>
+        </div>
+        <div class="copyright">
+          <p>© Cake by Mo Mahboobian, 2022. All rights reserved.</p>
+        </div>
+      </section>
+    </footer>
+    <script>
+      const hamburger = document.querySelector(".hamburger");
```
jonnywyatt, line 101: Nice!

jonnywyatt (commented):
Great job. You've clearly got a good eye for design as well as being able to code! 👏 
Well done on getting the burger menu working with Javascript!
There's not much to improve here, really!
One tip to improve page load performance would be to save your .jpg images with more compression so they're smaller files, will mean your mobile users will see the page quicker.

css/style.css
```diff
@@ -0,0 +1,276 @@
...
+  padding: 0;
+  box-sizing: border-box;
+}
+
+html {
+  font-family: "Roboto", sans-serif;
+  font-weight: 300;
+}
+
+/* ----- Header ----- */
```
jonnywyatt, line 14: Very well organised CSS, great that you're grouping sections under comments 👍

css/style.css
```diff
@@ -0,0 +1,276 @@
...
+  transform: translateY(8px) rotate(45deg);
+}
+
+.hamburger.active .bar:nth-child(3) {
+  transform: translateY(-8px) rotate(-45deg);
+}
+
+/* ----- Nav Menu ----- */
+.nav-menu {
+  position: fixed;
```
jonnywyatt: This works, but when you scroll the page on a small screen, the open menu covers the page contents behind it, due to `position: fixed`.
Alternative method: set `position: relative` on the `.header` container, and use `position:absolute` on the popup menu. Then the open menu will scroll upwards with the page.

PR #412, opened 2022-11-09
London9-Oleh_Pysmenko-HTML-CSS-Coursework-Week3

tiritaki:
live version https://tiritaki.github.io/HTML-CSS-Coursework-Week3/

# Your Details

- Your Name: Oleh Pysmenko
- Your City: London
- Your Slack Name:

# Homework Details

- Module:
- Week:3

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

-----
[View rendered readme.md](https://github.com/tiritaki/HTML-CSS-Coursework-Week3/blob/master/readme.md)


index.html
```diff
@@ -1,16 +1,61 @@
...
-    <meta charset="UTF-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <meta http-equiv="X-UA-Compatible" content="ie=edge">
-    <title>Responsive Cake webpage</title>
-    <!-- Add a link to your css file here -->
-</head>
+  <body>
+    <main class="homepage-main">
+      <section class="homepage-hero">
+        <div class="one-one">
```
nuclear-pasta, line 14: I think these are leftovers from our coding mob

index.html
```diff
@@ -1,16 +1,61 @@
...
+          <div></div>
+          <div></div>
+          <div></div>
+        </div>
+        <div class="one-two">The best cake in town <br>delivered to your door</div>
+        <div class="navigation">
```
nuclear-pasta, line 20: I would say, looking at the design that these belong to a second block without the background image

style.css
```diff
@@ -1 +1,296 @@
...
+     }
+     .phone {
+      grid-area: phone;
+      text-align: center;
+     }
+    .homepage-footer {
+      grid-area: footer;
+      display: grid;
+      
+      grid-template-columns: repeat(3, 1fr);
```
nuclear-pasta, line 154: I would give instead of a fraction, the exact amount in rem as second parameter. So that you can then use:
justify-content: 'center' and they will be centered as in the design. So to summarise
grid-template-columns: repeat(3, 6rem);
justify-content: center
"

style.css
```diff
@@ -1 +1,296 @@
...
+    .homepage-products {
+      grid-area: product;
+      display: grid;
+      
+      grid-template-columns: repeat(3, 1fr);
+      grid-template-areas:
+        "product-item1  product-item2  product-item3" ;          /* one for each column */ 
+    } 
+
+     .contact {
```
nuclear-pasta: I think you missed an s here. the class is "contacts"

style.css
```diff
@@ -1 +1,296 @@
...
+    justify-self: center;
+  }
+  
+  .homepage-footer {
+    background-color:rgb(194, 229, 241);
+    grid-area: footer;
+    padding-left: 15px;
+  }
+  .oleh {
+    text-align: center;
```
nuclear-pasta, line 101: if you use the previous suggestions:

grid-column: 1 / span 3; will center.
What is doing is taking the whole 3 spaces and then centering. Which seems more like the design

nuclear-pasta:
https://www.loom.com/share/52b25d4d74a349feb667a50b564e3bc2

index.html
tiritaki: I used this classes in my work :)

nuclear-pasta: what's their scope? I deleted them in DOM and could see no relevant changes

tiritaki: I used .one-one class to create hamburger sign in mobile version
Screenshot 2022-11-10 at 15 29 44

nuclear-pasta: got it

index.html
tiritaki: .navigation class I was using for desktop version. 
Only it goes inside first block, not as per design as separate block

style.css
tiritaki: thank you, good advice 
only I used space-around for justify-content
image

index.html
tiritaki: yes, you was right.
there were menu in tablet mode, I returned it :)

style.css
tiritaki: I was using grid-template-areas only for small cakes which are part of main grid.
I used contact in separate grid for footer
image

style.css
tiritaki: super, it helped to center it in all modes
image

nuclear-pasta: awesome. But the image you attached, still seems not totally centered. Probably is because of the class name called `contact` instead of `contacts`. Also in the design they seem to be all in the center compared to your image -> justify-content: center would help that
Screenshot 2022-11-10 at 17 03 28

tiritaki: yes, you are right 
about wrong class name for contacts. now it is working fine

tiritaki: image

tiritaki:
> https://www.loom.com/share/52b25d4d74a349feb667a50b564e3bc2

only in the evening I saw your video :) after fixing issues

PR #424, opened 2022-11-10
London9- pamela luna- HTML/CSS- WEEK 3- CAKE

pamelss:

# Your Details

- Your Name: Pamela Luna
- Your City: London
- Your Slack Name: Pamela Luna

# Homework Details

- Module: HTML-CSS
- Week: 3

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?


pamelss:
requesting to review my pull, thank you.

style.css
```diff
@@ -1 +1,247 @@
...
+
+.hero {
+    margin: 1em;
+}
+
+.hero h1 {
+    margin: 0 0 1rem 0;
+}
+
+.mainpicture {
```
ehwus, line 68: With class names / IDs in CSS, we generally use a hypen to separate words. This just makes it easier to read, so `mainpicture` would just become `main-picture`.

style.css
```diff
@@ -1 +1,247 @@
...
+    width: 100%;
+    height: 20rem;
+    object-fit: cover;
+}
+
+.products {
+    display: grid;
+    grid-template-columns: repeat(2, 1fr);
```
ehwus, line 76: You've identified a really good use case for fractional units here, great job!

style.css
```diff
@@ -1 +1,247 @@
...
+    padding: 0.6;
+    max-width: 13rem;
+    margin: 6rem 0 0 0;
+}
+
+.title p {
+    font-size: large;
+    color: rgb(246 6 202);
+    font-weight: 700;
+
```
ehwus, line 45: Let's just make sure we're eliminating white spaces like this before submitting PRs, it's a very small and nit picky thing but a good habit to get into.

style.css
```diff
@@ -1 +1,247 @@
...
+}
+
+.nav2 {
```
ehwus, line 48: Could this class name be a bit more descriptive? I don't immediately know what the 2 refers to here.

style.css
```diff
@@ -1 +1,247 @@
...
+    .hero h1{
+        font-size: 3rem;
+        font-family: 'cursiva';
+    }
+
+    .hero p {
+        font-size: larger;
+    }
+
+    .mainpicture{
```
ehwus, line 209: Again it's a minor formatting issue, but let's try to make sure that you're leaving a space between the selector name and the block consistently.

style.css
```diff
@@ -1 +1,247 @@
...
+
+
+
+
+
+
+
+
+
+}
```
ehwus, line 245: There's a lot of white space here, so I can't immediately see what this brace is actually closing. This is a good example of where poor formatting can affect the readability of the code.

I use autoformatting when I code, so every time it saves it automatically formats. I do however still try to largely write my code in a way that follows the coventions (it's a good habit to be in). This is a good guide on how to get that set up in VS Code https://www.alphr.com/auto-format-vs-code/

ehwus (commented):
You have met the requirements brief really well here Pam, there is a good alignment between the designs and your end result. It's hard work to make a fully responsive site, but your usage of grid was smart and has allowed you to keep the HTML simple (as it should be) and your CSS is clear.

The biggest issues here are very correctable - fixing your tooling will make these problems go away immediately. However, it's best not to rely entirely on tooling, and I would encourage you to have more attention to detail when it comes to the way your code is presented. You should be aiming to make your code as easy to read as possible, because then it's easier to maintain and add features.

I'd also follow conventions on naming of folders, and ideally just call your folder `images`. Having spaces in folder names online can cause complications, so it's best to avoid that wherever possible.

Overall, a really solid effort, you've demonstrated a strong foundation in creating responsive sites. It's lovely to see your creativity and use of colour coming through as well.

PR #438, opened 2022-11-11
LONDON CLASS_9-SAGHAR_HOSSEINMARDI-HTML-CSS-WEEK-3

SagharHosseinmardi:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?


index.html
```diff
@@ -1,16 +1,92 @@
...
+      <div class="header">
+        <img
+          src="design/bakery-scene-watercolor-cooking-illustration-kitchen-logo_257473-76-removebg-preview.png"
+          alt="cake logo"
+          class="logo-img"
+        />
+      </div>
+
+      <p class="header-text">
+        The best cakes in town <br />delivered to your door
```
jonnywyatt, line 27: Adding the line break with `br` means it's harder to change the wrapping at different screen widths. In the desktop design, the text should be on one line.
Instead, try controlling the line break using CSS by changing the width of the text container.

style.css
```diff
@@ -1 +1,96 @@
...
+}
+
+body {
+  margin: 0;
+  padding: 0;
+  font-family: "Lemonada", sans-serif;
+}
+
+.logo-img {
+  width: 30%;
```
jonnywyatt, line 15: Looks good, the image scales with the viewport width

style.css
```diff
@@ -1 +1,96 @@
...
+}
+
+header {
+  background-color: rgb(245, 225, 228);
+}
+
+.header-text {
+  float: right;
```
jonnywyatt, line 23: This does align the text correctly but there's a simpler way; set `display: flex` on the `header`, and then use `align-self: flex-end` on `.header-text` which will vertically align it to the bottom of the container.

style.css
```diff
@@ -1 +1,96 @@
...
+.item {
+  padding: 0.6em;
+}
+
+.copyright {
+  display: flex;
+  justify-content: center;
+}
+
+@media screen and (max-with: 420px) {
```
jonnywyatt, line 79: You don't need this media query block; the default styles will hide the navbar, then the media query below this one (for 900px) will show it.

style.css
```diff
@@ -1 +1,96 @@
...
+  height: 100px;
+  padding: 20px;
+  margin: 10px;
+}
+footer {
+  background-color: rgb(245, 225, 228);
+  color: rgb(86, 83, 83);
+}
+.footer-items {
+  display: flex;
```
jonnywyatt, line 66: These will stack vertically without needing to use flex, because `p` is a block element and they naturally occupy the whole width and stack one below the other.

jonnywyatt (commented):
Really good!

Clean HTML, well organised CSS. You're improving your skills with each project 👏

PR #442, opened 2022-11-11
London9-Elahe-Mortazavi-HTML-CSS-Coursework-Cake-Project-Week3

elahemortazavi:
first attempt to complete the cake-project-week3

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?


matthiastan13 (commented):
Nice design, I see you have a flair for it. 

Good experimentation with so many different hover effects, transitions and transformations.

Good work in making the page responsive. You managed to stack the links in the footer vertically, and also to put the welcome image and text on top of each other for mobile. You also added the hamburger icon for mobile. However, the 4 cake images should be arranged in a row for the desktop layout.

Still too much usage of px, should start learning to use rems now. This youtube [video](https://www.youtube.com/watch?v=okw-whFWGEo) might be quite helpful, and look around for others on youtube. You can use rems not just for fonts but also for things like margin and width etc.

You took some of your images from unsplash.com, that is good, because then you don't have to worry about copyright. It would be better to take all your images from there in future.

See some of the comments in the code.

index.html
```diff
@@ -6,11 +6,107 @@
...
-    <!-- Add a link to your css file here -->
-</head>
+     <!-- Add a link to your css file here -->
+    <link rel="stylesheet" href="./style.css">
+    </head>
 
 <body>
-    <!-- Add your markup here -->
+    <wrap class="container">
+
```
matthiastan13, line 15: There is no `wrap` element. See list of html elements on mdn [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element)

index.html
```diff
@@ -6,11 +6,107 @@
...
+                <ul class="header-ul">
+                    <li class="more-menu" id="show-product-menu">
+                        <a href="/index.html">HOME</a>
+                    </li>
+                    <li class="more-menu" id="show-product-menu">
+                        <a href="#">CAKES</a>
+                    </li>
+                    <li class="more-menu" id="show-product-menu">
+                        <a href="/level-2/store.html">ORDERIN</a>
+                    </li>
```
matthiastan13, line 47: This was taken from a previous assignment. Remember to remove the incorrect href. Id should also be used on only 1 element. If you want to label more than 1 element, use class instead.

index.html
```diff
@@ -6,11 +6,107 @@
...
+     
+     
+     
+     
+       
+       
+          <img class="main-cake" src="Images/pexels-shvets-production-7525182.jpg" alt="main-cake" />
+       
+          <fieldset class="welcome">
+            <h1 class="wel">Welcome</h1>
```
matthiastan13, line 67: Not correct use of `<fieldset>` here, just `div` should be ok. `fieldset` should be used to group labels or controls inside a `<form>`. See [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset)

style.css
```diff
@@ -1 +1,453 @@
...
+    "c3 c3 c3 c3 c3 c3 c4 c4 c4 c4 c4 c4"
+    "f f f f f f f f f f f f";
+    padding: 15px;
+ 
+}
+ 
+.social-ul {
+    display: grid;
+  flex-wrap: wrap;
+  justify-content:;
```
matthiastan13, line 207: flex-wrap and justify-content go with flexbox, i.e. `display:flex`, not css grid, i.e `display:grid`

PR #443, opened 2022-11-11
London class 9-lovelace-Mana Ghabraei-HTML/CSS cake

ghost:
It's initial version I'll complete that as soon as possible ;)


ghost:
done!

style.css
```diff
@@ -1 +1,357 @@
...
+}
+
+* {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
+
+body {
+  display: grid;
```
jonnywyatt, line 22: I don't think you need grid layout at this level. Remember, if you use block elements like `div` and `section`, they will naturally sit one below another and form a column, you don't need to add any CSS to achieve that.

style.css
```diff
@@ -1 +1,357 @@
...
+  font-family: "Montserrat", sans-serif;
+}
+/* header */
+header {
+  grid-area: 1 / 1 / 2 / 2;
+  background-color: var(pink);
+  /* header grid */
+  display: grid;
+  grid-template-columns: repeat(2, 1fr);
+  grid-template-rows: repeat(2, 1fr);
```
jonnywyatt, line 38: This does align the title at the bottom of the header correctly but there's a simpler way; set display: flex on the header, and then use align-self: flex-end on the `p` which will vertically align it to the bottom of the container.

index.html
```diff
@@ -5,12 +5,72 @@
...
 
 <body>
-    <!-- Add your markup here -->
+    <header>
+        <div class="logo">
+            <img src="Delicious-cupcake-PNG-1.png" alt="logo">
+            </div>
+            <nav>
+                <ul>
+                    <input src='checkbox' id='check' />
```
jonnywyatt, line 21: if you change `src` attribute to `type` it will behave like a checkbox

style.css
```diff
@@ -1 +1,357 @@
...
+    font-size: 24px;
+    text-align: right;
+  }
+  .mainpicture {
+    grid-area: 2 / 1 / 3 / 3;
+  }
+
+  .pic {
+    grid-area: 3 / 1 / 4 / 5;
+    grid-template-columns: repeat(4, 1fr);
```
jonnywyatt, line 339: This works well, the images snap into a single row 👍

jonnywyatt (commented):
Nice job! This works well responsively, the layout changes as the viewport width changes.

I think it's possible to do some things without writing any CSS eg HTML block elements like `div` and `section` will naturally flow in a column without any CSS, so usually no need to use `grid` or `flex` to get a column of items.

index.html
```diff
@@ -5,12 +5,72 @@
...
+                </div>
+                <label for="check" class="open-menu"><i class="fa-solid fa-bars"></i></label>
+            </ul>
+            </nav>
+            <p>Cupcakes & cakes Delivered to your door</p>
+    </header>
+    <main>
+        <section class="navigation">
+            <ul>
+                <li><a href="#">Home</a></li>
```
jonnywyatt, line 37: Not a big problem, but something to consider - you could avoid repeating all the nav items twice in the HTML, by moving the nav panel with CSS (on small screens, hide it then hover below the burger menu when clicked; on larger screens, show it below the header).

style.css
```diff
@@ -1 +1,357 @@
...
+  font-family: "Cookie", sans-serif;
+}
+
+.container p {
+  font-size: 30px;
+}
+
+.mainpicture {
+  grid-area: 2 / 1 / 3 / 2;
+  background-image: url(/image/Anniversary-Cake-Birthday-60th-Diamond-1024x731.jpeg);
```
jonnywyatt, line 153: This is probably a content image, so should be a `img` element in the HTML. 
`background-image` is good for decorative images like icons.

style.css
```diff
@@ -1 +1,357 @@
...
+footer {
+  grid-area: 3 / 1 / 4 / 2;
+  background-color: var(--yellow-light);
+  /* footer grid */
+  display: grid;
+  grid-template-columns: repeat(2, 1fr);
+  grid-template-rows: repeat(2, 1fr);
+  grid-column-gap: 0px;
+  grid-row-gap: 0px;
+  height: fit-content;
```
jonnywyatt, line 192: I don't think this property is needed, the footer will naturally grow to fit its content

index.html
ghost: oh, I'll chenged it as soon as possible thank you for your attention Jon

PR #445, opened 2022-11-11
 London Class9/Turing - Karla Grajales - HTML/CSS - Week 3

Grajales-K:

# Your Details

- Your Name: Karla Grajales
- Your City: London 9/Turing
- Your Slack Name: Karla Grajales

# Homework Details

- Module: HTML/CSS
- Week: 3

# Notes

- What did you find easy?
  flexbox

- What did you find hard?
Grid

Grid

- Any other notes?
Please find the project here.  https://cyf-grajales-k-cakes.netlify.app/


index.html
```diff
@@ -5,12 +5,78 @@
...
-    <title>Responsive Cake webpage</title>
-    <!-- Add a link to your css file here -->
+    <title>Responsive Cake Webpage</title>
+    <link
+    href="//fonts.googleapis.com/css?family=Roboto:400,500,300"
+    rel="stylesheet"
+    type="text/css"
+    />
+    <link rel="stylesheet" href="/style/style.css" />
+    <link rel="shortcut icon" type="image/x-icon" href="/img/red.jpg"/>
```
jonnywyatt, line 15: This works fine but it's better to use just `rel="icon"` ([ref](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types))

index.html
```diff
@@ -5,12 +5,78 @@
...
+
 </head>
 
 <body>
-    <!-- Add your markup here -->
-</body>
+    <header class="header">
+        <!-- <img src="/img/Cupcake.png" alt="" class="header-pic"> -->
+        <nav class="nav-header">
+            <li href="#">HOME</li>
```
jonnywyatt, line 24: The list of `li` needs to be wrapped in a `ul` or `ol` tag

index.html
```diff
@@ -5,12 +5,78 @@
...
+                <li href="#">Ordering</li>
+                <li href="#">LESSON</li>
+                <li href="#">ABOUT</li>
+            </ul>    
+        </nav>
+    </div>
 
-</html>
\ No newline at end of file
+    <section class="homepage-main">
```
jonnywyatt, line 49: This would be good as a `main` tag. A page should only have one `main` element and it should wrap the dominant content area ([ref](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main))

index.html
```diff
@@ -5,12 +5,78 @@
...
+            </ul>    
+        </nav>
+    </div>
 
-</html>
\ No newline at end of file
+    <section class="homepage-main"> 
+        <div class="main-hero">
+            <h1 class="title">Welcome</h1>
+            <p class="text-hero">It’s really not a special occasion without a generous slice of cake or two.
```
jonnywyatt, line 52: Wow you really know how to sell cakes! 😁

index.html
```diff
@@ -5,12 +5,78 @@
...
+            </p>
+        </div>
+        <div class="wrapper-img">
+            <div class="img-hero" >
+                <img src="/img/strawberrry cake.jpg" alt="wedding strawberry cake">
+            </div>
+        </div>
+        <div class="products">  
+            <div class="homepage-products">
+                <img src="/img/wedding.avif" alt="wedding cake" class="wedding">
```
jonnywyatt, line 67: This isn't a big deal for this exercise, but just including the info out of interest - 
.avif image format isn't supported by all main browsers; .webp might be a better choice as it also has very good small file sizes and wide browser support.

index.html
```diff
@@ -5,12 +5,78 @@
...
+                <img src="/img/sweet.avif" alt="mini strawberrry cakes" class="sweet">
+            </div>
+        </div>      
+    </section>
+
+    <footer>
+        <span class="budget">Make your budget with us</span>
+        <span class="addres">Threadneedle Street, London, EC2R 8A Enquiries: +44(0)20 3461 4878</span>
+        <span class="email" type="email">email</span>
+        <small class="copy">&#169;Karla Grajales</small>
```
jonnywyatt, line 79: `small` is a good choice of semantic tag for a copyright line 👏

style/style.css
```diff
@@ -0,0 +1,312 @@
...
+*{
+    margin: 0;
+    padding: 0;
+    box-sizing: border-box;
+    font-family: "Roboto", sans-serif;
+}
+
+body{
+    background-color: #ffffff;
+    display: grid;
```
jonnywyatt, line 11: This works fine, but don't forget, HTML 'block' elements like `header`, `div` and `section` will naturally, without any CSS, stack vertically one below the other, and fill the available width. They don't need to be 'held in place' with a grid at `body` level, if that makes sense? eg you could set `min-height: 200px` on the `header` and a bit of `margin-bottom`, and you would achieve the same layout with less CSS. Having said that, it's not really a problem that you've used grid here.

style/style.css
```diff
@@ -0,0 +1,312 @@
...
+    "footer";
+    gap: 1rem;
+}
+
+
+.header{
+    /* background-color: red; */
+    grid-area: hd;
+    background-image: url("/img/header\ pic.avif");
+    object-fit: cover;
```
jonnywyatt, line 26: `object-fit` is for `img` tags; you've used a background image here so I don't think you need this property. Try unticking it in devtools Styles panel and see if the layout changes.

style/style.css
```diff
@@ -0,0 +1,312 @@
...
+    background-size: cover;
+    background-repeat: no-repeat;
+    background-position: center;
+    display: flex;
+    justify-content: flex-end;
+    align-items: flex-end;
```
jonnywyatt, line 32: Nicely done!

style/style.css
```diff
@@ -0,0 +1,312 @@
...
+    width: 100%;
+}
+
+.nav-header {
+    display: none;
+}
+
+.add {
+    max-width: 24rem;
+    word-wrap: break-word;
```
jonnywyatt, line 42: I don't think you need this; it will cause fragments of words to break to the next line. Best for readability to let words stay whole and wrap using the whitespace between them.

style/style.css
```diff
@@ -0,0 +1,312 @@
...
+
+}
+
+.add-text {
+    color: rgb(255, 255, 255);
```
jonnywyatt, line 47: Putting rgb colour values inline like this works fine 👍
Even better is to use CSS variables to store the colours, will make them readable as you use them in your CSS. You can declare them once in a :root selector then re-use them ([example](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties#using_the_root_pseudo-class))

style/style.css
```diff
@@ -0,0 +1,312 @@
...
+    text-shadow: 0 0 20px #dae351, 0 0 10px #ff0000;
+    font-family: fantasy;
+    text-align: right;
+    font-size: 2rem;
```
jonnywyatt, line 51: Great 👏 
rem's are better than px for accessibility. If the user resizes text in their browser settings, this font size will respond.
Base font size in most browsers is 16px, so 2rem === 32px

style/style.css
```diff
@@ -0,0 +1,312 @@
...
+.add {
+    max-width: 24rem;
+    word-wrap: break-word;
+
+}
+
+.add-text {
+    color: rgb(255, 255, 255);
+    text-shadow: 0 0 20px #dae351, 0 0 10px #ff0000;
+    font-family: fantasy;
```
jonnywyatt, line 49: It's fine to use custom fonts, but try to specify a fallback font as well, in case the custom font isn't available on the user's device.
eg `font-family: fantasy, Serif`
([more info](https://css-tricks.com/css-basics-fallback-font-stacks-robust-web-typography/))

style/style.css
```diff
@@ -0,0 +1,312 @@
...
+        margin-top: 1rem;
+        font-size: 1.5rem;
+        text-align: right;
+        font-weight: 900;
+    }
+
+    .text-hero{
+        text-align: right;
+        font-size: smaller;
+        padding: auto;
```
jonnywyatt, line 263: I don't think this is valid. You can see styles that are invalid / not used by looking at Chrome devtools, in the Elements panel select this element and look in the Styles panel
image

jonnywyatt (commented):
Really good! Looks great, good responsive layout. 🎉

PR #449, opened 2022-11-11
London 9 - Lovelace - Sahar Naderi - HTML/CSS - week 3

SaharNaderi:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?


index.html
```diff
@@ -1,16 +1,77 @@
...
-    <!-- Add a link to your css file here -->
-</head>
-
-<body>
-    <!-- Add your markup here -->
-</body>
-
-</html>
\ No newline at end of file
+        <link rel="icon" href="cup-cake-icon.png" type="image/icon type">
```
ehwus, line 3: Be careful here - you've got rid of the HTML tag with the DOCTYPE declaration. This can cause all sorts of problems, particularly with older browsers. Make sure that when you're changing code you're not accidentally deleting something that's important.

index.html
```diff
@@ -1,16 +1,77 @@
...
+                    </label>
+                </ul>
+            </nav>
+            <h1>The best cakes in the world for delivered to you</h1>
+        </header>
+      
+    
+        <main>
+    
+            <section class="navigation">
```
ehwus, line 37: It's best to avoid empty sections like this if you can. I appreciate that sometimes it's necessary to use something like this to display an image with CSS (like you've done really well with the hero-pic below), but I can't see the use case for this one in particular.

index.html
```diff
@@ -1,16 +1,77 @@
...
+        <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.2.0/css/all.min.css" integrity="sha512-xh6O/CkQoPOWDdYTDqeRdPCVd1SpvCA9XXcUnZS2FmJNp1coAFzvtCN9BmamE+4aHK8yyUHUSCcJHgXloTyT2A==" crossorigin="anonymous" referrerpolicy="no-referrer" />
+        <title>Responsive Cake webpage</title>
+        <link rel="stylesheet" href="style.css">
+    </head>
+    
+    <body>
+    
+        <header>
+            <div class="logo">
+                <img src="piccake/IMG_1785 2.PNG" alt="logo">
```
ehwus, line 13: Lovely that you're keeping the habit of adding alt text to all of your images!

style.css
```diff
@@ -1 +1,319 @@
...
+    display: grid;
+    grid-template-columns: 1fr;
+    grid-template-rows: repeat(3, 1fr);
+    grid-column-gap: 0px;
+    grid-row-gap: 0px;
+}
+
+/* -----header------ */
+
+header {
```
ehwus, line 21: This is a really good example of putting together a grid, nice work!

style.css
```diff
@@ -1 +1,319 @@
...
+}
+
+nav {
+    grid-area: 1 / 2 / 2 / 3;
+}
+
+.menu {
+    display: flex;
+    flex-direction: column;
+    position: absolute;
```
ehwus, line 51: Did the menu need to be an absolute position here? You have set up the main page to use a grid system, so it would be best to continue in this style throughout the exercise. There's a good article on using a full grid to layout a page here: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/

style.css
```diff
@@ -1 +1,319 @@
...
+    top: -600px;
+    right: 120px;
+    transition: all .5s ease-in-out;
```
ehwus, line 54: Great to see use of transition here!

ehwus (commented):
The objective of the exercise was to demonstrate understanding of the grid and responsive design, and I think you've done that well. I can see though that you've had a bit of a hard time organising the elements within the page themselves - grid is really hard on it's own, and you've opted for one of the hardest things in arranging literally everything in a grid format. 

I've worked as a specialist frontend engineer at places that avoid the use of grid in this way because it's so hard to do, so I think with that in mind you've done a really good job. I think next time you could simplify the task by thinking about the page in a normal CSS flow, with smaller grids for individual parts of the site. For instance, the images could be a grid.

I hope that makes sense, and I'm around all day if you want to talk about this, but overall you've done a really good job on navigating one of the hardest parts of CSS! I've left a couple of specific comments for things that you should watch out for, and there are also general formatting improvements that would be a good idea to automate (https://www.alphr.com/auto-format-vs-code/)

PR #460, opened 2022-12-05
Emiy_Barton_HTML/CSS_week_3

EABarton123:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/index.html
+++ b/index.html
@@ -16,18 +16,10 @@
     <!-- Add your markup here -->
     <header class="nav">
         <nav class="nav-bar">
-            <ul>
-                Home
-            </ul>
-            <ul>
-                About us
-            </ul>
-            <ul>
-                Price List
-            </ul>
-            <ul>
-                Contact us
-            </ul>
+            <a>Home</a>
+            <a>About Us</a>
+            <a>Price List</a>
+            <a>Contact us</a>
         </nav>
     </header>
     <main>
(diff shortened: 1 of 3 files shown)


style.css
```diff
@@ -1 +1,56 @@
 /*  Add your styling here */
+body  {
+    display: grid;
+    grid-template-areas: 1fr 2fr 4fr 3r;
```
SallyMcGrath:
```
    grid-template-areas: 1fr 2fr 4fr 3r;
```
What is this rule trying to do here? Can you take another look?

index.html
```diff
@@ -6,11 +6,62 @@
...
+    <link rel="stylesheet" href="style.css" />
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
     <!-- Add a link to your css file here -->
 </head>
 
 <body>
     <!-- Add your markup here -->
+    <header class="nav">
+        <nav class="nav-bar">
```
SallyMcGrath, line 18: When you validated this HTML, what result did you get?

What is the meaning of `<ul>` ? Is there another piece to this code that might help you here? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul

index.html
```diff
@@ -6,11 +6,62 @@
...
+            </h1>
+            <h2 class="tagline">
+                Here at the Vegan Cake shop we pride ourselves on tasty cakes that use 100% organic and vegan ingredients. We believe cakes do not need to upset the environment or harm any animals in the process. See our price list for all our pre-made delicious cakes or contact us for bespoke requests.
+            </h2>
+            <button class="order-now">
+                Order Now
+            </button>
+        </section>
+        <section class="cake">
+            <figure class="img1">
```
SallyMcGrath: A class typically means a group of things. It's a way to set rules at a group level, instead setting rules element by element.

Is there a class you think of that could apply to _all_ of these figures?

index.html
```diff
@@ -6,11 +6,62 @@
...
+                <img src="https://cdn.shopify.com/s/files/1/2725/9456/products/flavourtown-bakery-london-cakes_vegan-rainbow-new-01_2048x2048.jpg?v=1613475666" alt="picture of a vegan rainbow cake" width="25%">
```
SallyMcGrath: Great alt text here. Really good work. 🏅 

You can make this alt text even better by removing the "picture of" wording. A screenreader will announce it is a picture anyway.

index.html
```diff
@@ -6,11 +6,62 @@
...
+            <figure class="img3">
+                <img src="https://cupfulofkale.com/wp-content/uploads/2018/10/whole-cake-735x1102.jpg" alt="picture of a vegan elderflower naked cake" width="25%">
+            </figure>
+            <figure class="img4">
+                <img src="https://eluxemagazine.com/wp-content/uploads/2018/04/unicorn_birthday_cake_recipe.jpg" alt="picture of a vegan unicorn cake" width="25%">
+            </figure>
+        </section>
+        </main>
+        <footer class="social">
+            <a href="https://image.shutterstock.com/image-photo/image-260nw-278925056.jpg"></a>
```
SallyMcGrath:
```suggestion
            <a href="https://www.facebook.com/"><img src="https://image.shutterstock.com/image-photo/image-260nw-278925056.jpg" alt="facebook" /></a>
```

I don't think these links could show anything the way you have them. What do you think?

style.css
```diff
@@ -1 +1,56 @@
...
+    padding-bottom: 2em;
+    font-size: 1em;
+}
+
+.info {
+    text-align: center;
+    background-color: #ff80ff;
+}
+
+.cake-images {
```
SallyMcGrath:
```suggestion
.cake {
```

I don't think this rule can be applying as there's no cake-images class in the HTML that I can see. Devtools should help you find and fix these errors.

https://developer.chrome.com/docs/devtools/

style.css
```diff
@@ -1 +1,56 @@
 /*  Add your styling here */
+body  {
+    display: grid;
+    grid-template-areas: 1fr 2fr 4fr 3r;
+    background-color: #ffe6ff;
+}
+
+.nav-bar > ul {
+   display: inline;
+   padding: 100px 6em 6em 7em;
```
SallyMcGrath: I suggest using rems instead of ems in general. 

Ems are a relational scaling element that can create huge problems when users scale up their text. Lots of users do this.

style.css
```diff
@@ -1 +1,56 @@
...
+    text-align: center;
+    background-color: #ff80ff;
+}
+
+.cake-images {
+    display: flex;
+}
+
+.img1 {
+    width: 25%;
```
SallyMcGrath: Instead of setting 25% on here (and you've also set it on the HTML), consider instead using flex-basis on the figure elements.

Then size your images to be 100% of whatever that containing figure element is.

SallyMcGrath (commented):
Thanks for opening this PR, @EABarton123 !! 🎉 

Really appreciate the work you've done so far. I think it might be worth you taking another look at this project to see what you can improve.

style.css
EABarton123: thank you for this, I have removed it as I hadnt actually done anything with the grid and I should have names areas here.

index.html
EABarton123: I have changed to <a> think this is better?

index.html
EABarton123: i have removed the class and just used figure

index.html
EABarton123: thank you

index.html
EABarton123: yep thank you

PR #497, opened 2023-02-08
London10 - Khalil Alhaydr -HTML-CSS-Week3-Cake Co

Khlil1313:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?


index.html
```diff
@@ -7,10 +7,68 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Responsive Cake webpage</title>
     <!-- Add a link to your css file here -->
+    <link rel="stylesheet" type="text/css" href="style.css" />
+    <script type="text/javascript">
```
jxz12, line 11: this is a good attempt

index.html
```diff
@@ -7,10 +7,68 @@
...
+        function classToggle() {
+        const navs = document.querySelectorAll('.navbar_items')
```
jxz12, line 13: I would indent these two lines to the right to make it look like they are separate

index.html
jxz12: The script is getting called too early, which is why it cannot find the <div> with class `.navbar_links_toggle`

jxz12: `this looks like code`

